PR #19, opened 2023-06-07
fix: bff 724/screens

alexbejann:
## Type of Change
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/core/network/api_service.dart
```diff
@@ -127,4 +127,26 @@ class APIService {
...
+        'Content-Type': 'application/json',
+      },
+      encoding: Encoding.getByName('utf-8'),
+      body: jsonEncode(body),
+    )
+        .then((response) {
+      if (response.statusCode >= 400) {
+        throw Exception('something went wrong :(');
+      }
+      return response.statusCode == 200;
```
MaikelStuivenberg, line 149: Is there a specific reasons why only a 200 is true? Or should that always return true if its not >400

lib/features/give/bloc/give/give_bloc.dart
```diff
@@ -0,0 +1,150 @@
...
+    GiveQRCodeScanned event,
+    Emitter<GiveState> emit,
+  ) async {
+    emit(state.copyWith(status: GiveStatus.loading));
+    try {
+      final uri = Uri.parse(event.rawValue);
+      final mediumId = utf8.decode(base64.decode(uri.queryParameters['code']!));
+      final organisation = await _getOrganisation(mediumId);
+      final transactionList = _createTransationList(mediumId, event.userGUID);
+      //   'advertisementImageUrl': '',
```
MaikelStuivenberg, line 40: This comment block can be removed, right? 
Its also as code in a other file :)

lib/features/give/models/webview_input.dart
```diff
@@ -0,0 +1,228 @@
...
+    areYouSureToCancelGivts,
+    advertisementTitle,
+    apiUrl,
+    givtObj,
+  ];
+  
+}
+
+
+// 'advertisementImageUrl': '',
```
MaikelStuivenberg, line 204: remove comment block

lib/features/give/pages/organization_list_page.dart
```diff
@@ -0,0 +1,264 @@
...
+            icon: 'assets/images/campaign.png',
+            activeIcon: 'assets/images/campaign_focus.png',
+            color: AppTheme.givtOrange,
+            onTap: () => bloc.add(
+              OrganisationTypeChanged(
+                CollecGroupType.campaign.value,
+              ),
+            ),
+          ),
+          FilterSuggestionCard(
```
MaikelStuivenberg, line 220: I don't think that Artists is visible right now as a filter

MaikelStuivenberg (changes requested):

lib/core/network/api_service.dart
alexbejann: Well , the android app is only using the status code. If the status code is >=300 is retrying to send them 3 times. I am assuming that's because of the internet connection. Also, before it's caching the givts and on 200 is removing them

lib/features/give/pages/organization_list_page.dart
alexbejann: #18 has this change

lib/features/give/bloc/give/give_bloc.dart
alexbejann: #18 is addressing this

MaikelStuivenberg (approved):

PR #24, opened 2023-06-09
feat: bff 622/bt beacons

alexbejann:
## Description

Implemented the beacons part for testing without the last donation

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/features/give/bloc/give/give_bloc.dart
```diff
@@ -106,24 +103,134 @@ class GiveBloc extends Bloc<GiveEvent, GiveState> {
...
+    }
+    emit(state.copyWith(status: GiveStatus.processingBeaconData));
+    try {
+      final hex = StringBuffer();
+      for (final b in event.serviceData[event.serviceUUID]!) {
+        hex.write(sprintf('%02x', [b]));
+      }
+
+      final beaconData = hex.toString();
+      final contains = beaconData.contains('61f7ed01') ||
```
pavelpavelko, line 149: I think it would be helpful for further maintenance to add some brief explanation of the parsing process, since such vars like 'contains' or 'a' are not very self-explanatory :)

lib/features/give/pages/bt_scan_page.dart
```diff
@@ -0,0 +1,185 @@
...
+      }
+      if (scanResult.advertisementData.serviceUuids.isEmpty) {
+        continue;
+      }
+      if (!scanResult.advertisementData.serviceData
+          .containsKey(scanResult.advertisementData.serviceUuids.first)) {
+        continue;
+      }
+
+      if (scanResult.rssi < -75 || scanResult.rssi < -67) {
```
pavelpavelko, line 62: same here with values -75 and -67, what do they mean? Shall it be constants with clear names instead?

lib/features/give/pages/bt_scan_page.dart
```diff
@@ -0,0 +1,185 @@
...
+                  child: Image.asset('assets/images/givt_animation.gif'),
+                ),
+                Expanded(child: Container()),
+                Visibility(
+                  visible: _isVisible,
+                  child: Padding(
+                    padding: const EdgeInsets.all(20),
+                    child: ElevatedButton(
+                      onPressed: () => Navigator.of(context).push(
+                        MaterialPageRoute<void>(
```
pavelpavelko, line 132: maybe this is bc of lack of my bloc experience, but I do not understand the code in the push method

pavelpavelko (approved):
lgtm) 
I did not do a deep review, however code looks good and clear (except few places I commented)

lib/features/give/bloc/give/give_bloc.dart
alexbejann: We have internal documentation for those variables

lib/features/give/pages/bt_scan_page.dart
alexbejann: If the bloc is not available in other screens, where different BuildContext is created. Unless the BLoC is added at the MaterialApp widgets, as you have it in the Kids app. So to make it available in other screen i have to provide it to make it accessible. When i pop the screen the bloc will be automatically disposed instead of having a reset method and have it globally available.

lib/features/give/pages/bt_scan_page.dart
alexbejann: [rssi](https://bluetoothle.wiki/rssi) is the power of the frequency of the bluetooth device. If the device is further away meaning that is -57 or -40 for example we are skipping it.

PR #145, opened 2023-07-26
fix: Bff 646 download yearly overview

Daniela510:
## Description

Don't merge yet!  
I need to add github env variables for aws base api.

This PR switches to the newer AWS endpoint. 
Now the yearly summary comes as a separate PDF in email.
Also reverted the overview page to being state-less.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -27,6 +27,8 @@ jobs:
             echo "AMPLITUDE_KEY=f4fa9ab88de04c56a346aaa36c172a9a" >> $GITHUB_ENV
             echo "API_URL_US=backend.givt.app" >> $GITHUB_ENV
             echo "API_URL_EU=backend.givtapp.net" >> $GITHUB_ENV
+            echo "API_URL_AWS_US=api.production.givt.app" >> $GITHUB_ENV
+            echo "API_URL_AWS_EU=api.production.givtapp.net" >> $GITHUB_ENV
           else
             echo "Running on a different branch, deploying to test."
             echo "PACKAGE_NAME=net.givtapp.droid2.test" >> $GITHUB_ENV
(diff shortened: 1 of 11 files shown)


lib/features/overview/bloc/givt_bloc.dart
```diff
@@ -233,9 +233,11 @@ class GivtBloc extends Bloc<GivtEvent, GivtState> {
     Emitter<GivtState> emit,
   ) async {
     try {
+      final fromDate = '${event.year}-01-01';
+      final tillDate = '${event.year}-12-31';
+
```
alexbejann, line 238: Maybe is better to use a DateTime object and a DateFormatter

lib/features/overview/bloc/givt_bloc.dart
```diff
@@ -244,14 +246,10 @@ class GivtBloc extends Bloc<GivtEvent, GivtState> {
...
-      await LoggingInfo.instance.error(
-        e.toString(),
-        methodName: StackTrace.current.toString(),
-      );
-      emit(GivtError(
-        e.toString(),
-      ));
+    } on GivtServerFailure catch (e) {
+      await LoggingInfo.instance.info(e.body.toString());
+      await LoggingInfo.instance.error(e.body.toString());
```
alexbejann, line 251: Do you need both?

.github/workflows/android.yml
```diff
@@ -27,6 +27,8 @@ jobs:
             echo "AMPLITUDE_KEY=f4fa9ab88de04c56a346aaa36c172a9a" >> $GITHUB_ENV
             echo "API_URL_US=backend.givt.app" >> $GITHUB_ENV
             echo "API_URL_EU=backend.givtapp.net" >> $GITHUB_ENV
+            echo "API_URL_AWS_US=api.production.givt.app" >> $GITHUB_ENV
+            echo "API_URL_AWS_EU=api.production.givtapp.net" >> $GITHUB_ENV
```
alexbejann, line 31: This would have to be cleaned once we get rid of the AWS stuff

alexbejann (approved):
Just some remarks, but it's good enough

lib/features/overview/bloc/givt_bloc.dart
Daniela510: To use DateTime and a DateFormatter the lines would look like:
```
final fromDate = DateFormat('yyyy-MM-dd').format(DateTime.utc(int.parse(year), 01, 01));
final tillDate = DateFormat('yyyy-MM-dd').format(DateTime.utc(int.parse(year), 12, 31));
```
I dont really see the point of parsing the year string into and integer, to then create a dateTime object, to then format that into a string again. 
What are the benefits of using DateTime?

lib/features/overview/bloc/givt_bloc.dart
Daniela510: Nope

PR #157, opened 2023-08-01
Feat: bff 649 Personal summary per month

Daniela510:
## Description

New Feature, see monthly giving total and history.

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: equal_keys_in_map
+
 import 'dart:convert';
 import 'dart:io';
 
@@ -365,5 +367,32 @@ class APIService {
         body: jsonDecode(response.body) as Map<String, dynamic>,
       );
     }
+  }
+
+  Future<List<dynamic>> fetchMonthlySummary(
+    String guid,
+    String fromDate,
+    String tillDate,
+  ) async {
+    final params = {
+      'OrderType': '3',
+      'GroupType': '2',
+      'FromDate': fromDate,
+      'TillDate': tillDate,
+      'TransactionStatusses': '1',
+      'TransactionStatusses': '2',
+      'TransactionStatusses': '3',
+    };
+    final url = Uri.https(apiURL, '/api/v2/users/$guid/summary', params);
+
+    final response = await client.get(url);
+
+    if (response.statusCode >= 400) {
+      throw GivtServerFailure(
+        statusCode: response.statusCode,
+        body: jsonDecode(response.body) as Map<String, dynamic>,
+      );
+    }
+    return jsonDecode(response.body) as List<dynamic>;
   }
 }
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -4,7 +4,7 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:givt_app/core/failures/failures.dart';
-import 'package:givt_app/core/network/interceptor.dart';
+import 'package:givt_app/core/network/network.dart';
 import 'package:http/http.dart';
 import 'package:http_interceptor/http/http.dart';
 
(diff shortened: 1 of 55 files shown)


lib/features/personal_summary/pages/personal_summary_page.dart
```diff
@@ -0,0 +1,440 @@
...
+                    ),
+                    const Text(
+                      //locals.budgetSummarySetGoal,
+                      'Giving goal feature coming soon!',
+                      textAlign: TextAlign.center,
+                      style: TextStyle(
+                        fontSize: 14,
+                      ),
+                    ),
+                    const Spacer(),
```
alexbejann, line 239: Did you test with the spacer on a smaller device?

lib/shared/repositories/givt_repository.dart
```diff
@@ -115,4 +122,17 @@ class GivtRepositoryImpl with GivtRepository {
...
   }
+
+  @override
+  Future<List<MonthlySummaryItem>> fetchMonthlySummary({
+    required String guid,
+    required String fromDate,
+    required String tillDate,
+  }) async {
+    final decodedJson =
+        await apiClient.fetchMonthlySummary(guid, fromDate, tillDate);
```
alexbejann, line 133: I would make this one a bit more abstract, like a map to avoid conflicts when we would split that file

lib/utils/util.dart
```diff
@@ -55,6 +55,18 @@ class Util {
...
+    if (country == Country.us) {
+      currency = r'$';
+    }
+    if (Country.unitedKingdomCodes().contains(country.countryCode)) {
+      currency = '£';
+    }
+
+    return currency;
+  }
+
```
alexbejann, line 69: I wouldn't do this. Me and Pavlo use the NumberFormat to get the symbol

lib/core/network/api_service.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: equal_keys_in_map
```
alexbejann, line 1: Why?

alexbejann (approved):
Just address the comments

lib/features/personal_summary/pages/personal_summary_page.dart
```diff
@@ -0,0 +1,440 @@
...
+                    Text(
+                      locals.budgetSummarySetGoalBold,
+                      style: const TextStyle(
+                        fontSize: 14,
+                        fontWeight: FontWeight.w900,
+                      ),
+                    ),
+                    const Text(
+                      //locals.budgetSummarySetGoal,
+                      'Giving goal feature coming soon!',
```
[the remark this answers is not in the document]
Daniela510: The commented out one is the correct string for that widget, but since the button is inactive i wrote that temporarily to not confuse users. As soon as its implemented will be changed to the l10n string.

lib/features/personal_summary/pages/personal_summary_page.dart
Daniela510: No i thought spacer would default to zero then. But i changed this to exclude them.

lib/utils/util.dart
Daniela510: Thanks for the pointer ! 💪

lib/core/network/api_service.dart
Daniela510: This moved to givt_repository.dart. Its because of the way the backend expects the params.
```
final params = {
      'OrderType': '3',
      'GroupType': '2',
      'FromDate': fromDate,
      'TillDate': tillDate,
      'TransactionStatusses': '1',
      'TransactionStatusses': '2',
      'TransactionStatusses': '3',
    };
```
It does not take a list on TransactionStatusses, so there have to be equal keys in the map to pass multiple statusses (Entered, to be processed, processed). If i don't add the linter marks the file with a warning and its very distracting & annoying. Its not a hard need to have that, but a strong want 🙏

PR #187, opened 2023-08-21
feat: KIDS 918 register payment details US - Stripe

Daniela510:
## Description

Expanded the user registration to include US, and stripe as a PSP.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/features/registration/bloc/registration_event.dart
```diff
@@ -88,6 +88,20 @@ class RegistrationSignMandate extends RegistrationEvent {
   List<Object> get props => [guid, appLanguage];
 }
 
+class RegistrationStripeInit extends RegistrationEvent {
+  const RegistrationStripeInit();
+
+  @override
```
pavelpavelko, line 94: minor - why do you need to override props if you do not add any new members in the class?

lib/features/registration/bloc/registration_event.dart
```diff
@@ -88,6 +88,20 @@ class RegistrationSignMandate extends RegistrationEvent {
...
+  List<Object> get props => [];
+}
+
+class RegistrationStripeSuccess extends RegistrationEvent {
+  const RegistrationStripeSuccess();
+
+  @override
```
pavelpavelko, line 101: same here

lib/features/registration/pages/personal_info_page.dart
```diff
@@ -354,12 +400,13 @@ class _PersonalInfoPageState extends State<PersonalInfoPage> {
     });
     context.read<RegistrationBloc>().add(
           RegistrationPersonalInfoSubmitted(
-            address: _address.text,
-            city: _city.text,
-            postalCode: _postalCode.text,
+            address: _address.text.isEmpty ? 'Foobarstraat 5' : _address.text,
```
pavelpavelko, line 403: not sure why do we need these 'Foo' values? shall they be constants instead?

lib/features/registration/pages/signup_page.dart
```diff
@@ -40,228 +44,257 @@ class _SignUpPageState extends State<SignUpPage> {
...
+                if (state.status == RegistrationStatus.createStripeAccount) {
+                  context.read<StripeCubit>().fetchStripeCustomerCreationURLs();
+                }
+              },
+              builder: (context, state) {
+                return (state.status ==
+                            RegistrationStatus.createStripeAccount ||
+                        stripestate.stripeStatus == StripeObjectStatus.success)
+                    ? _createView(stripeCubit.state.stripeStatus)
+                    : Form(
```
pavelpavelko, line 142: I think this build method can be splitted on smaller pieces somehow for better readability? Maybe the Form can be a separate widget or function? but it is minor

pavelpavelko (approved):
lgtm with few minor comments :)

lib/features/registration/pages/signup_page.dart
Daniela510: Great point! It's done and it does help, especially because of all the conditional rendering on that page 😅

lib/features/registration/bloc/registration_event.dart
Daniela510: Why? - Because I did not question it, i removed it now. 🙏

PR #208, opened 2023-09-04
KIDS 1241 Hide family item

Daniela510:
## Description

Hide family options for current US users since functionality is not fully implemented.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/lib/shared/widgets/custom_navigation_drawer.dart
+++ b/lib/shared/widgets/custom_navigation_drawer.dart
@@ -235,7 +235,7 @@ class CustomNavigationDrawer extends StatelessWidget {
           ),
           if (showFamilyItem) _buildEmptySpace(),
           _buildMenuItem(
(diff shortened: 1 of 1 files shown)


lib/shared/widgets/custom_navigation_drawer.dart
```diff
@@ -235,7 +235,7 @@ class CustomNavigationDrawer extends StatelessWidget {
           ),
           if (showFamilyItem) _buildEmptySpace(),
           _buildMenuItem(
-            isVisible: showFamilyItem,
+            isVisible: false,
```
MaikelStuivenberg, line 238: So we still want to show it on dev environment?

Daniela510: We do, fixed.

PR #225, opened 2023-09-08
KIDS 1268 Add amplitude tracking to family features

Daniela510:
## Description

Enabling the amplitude tracking of the family features.

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


.vscode/launch.json
```diff
@@ -39,7 +39,7 @@
         "--target", 
         "lib/main_staging.dart",
         "--dart-define",
-        "AMPLITUDE_KEY=ceb9aaa139ac6028aa34166d6f57923e",
+        "AMPLITUDE_KEY=f4fa9ab88de04c56a346aaa36c172a9a",
```
alexbejann, line 42: This shouldn't have been added to the secrets?

lib/core/enums/amplitude_events.dart
```diff
@@ -1,6 +1,17 @@
 enum AmplitudeEvents {
```
alexbejann, line 1: This is not gonna be scalable. We have to come up with a better more generic solution. If i would create an event for every feature it's gonna be a huge amount of work to keep track of these events

lib/features/children/overview/pages/children_overview_page.dart
```diff
@@ -35,7 +36,13 @@ class ChildrenOverviewPage extends StatelessWidget {
           children: [
             Row(
               children: [
-                BackButton(),
+                BackButton(
+                  onPressed: () {
+                    Navigator.maybePop(context);
```
alexbejann, line 41: Why are you using the Navigator class? As we use the go_router for this? Like "context.pop()"

.vscode/launch.json
Daniela510: They were already commited in git, but prod vs dev was switched.

lib/core/enums/amplitude_events.dart
Daniela510: Do you mean this from an flutter implementation perspective? A solution being some sort of link between flutter and amplitude so events planned in amplitude can be accessed without defining them. Like locals?
Or from a broader perspective to use more general events in amplitude as well?

alexbejann: Yes, this way of doing it is not scalable. I was thinking about general events. Using locals is not a good idea as that's meant for translations anyway. Or a framework that would allow us to not implement/add so many custom events in a file but rather.

alexbejann (approved):

PR #226, opened 2023-09-09
feat: bff 653/discover

alexbejann:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/features/give/widgets/choose_category.dart
```diff
@@ -0,0 +1,196 @@
...
+                child: FilterSuggestionCard(
+                  isFocused: false,
+                  color: CollecGroupType.charities.color,
+                  title: _getCategoryName(
+                    type: CollecGroupType.charities,
+                    context: context,
+                  ),
+                  activeIcon: CollecGroupType.charities.activeIcon,
+                  icon: CollecGroupType.charities.icon,
+                  width: size.width * 0.4,
```
MaikelStuivenberg, line 68: Not sure if this was existing code, but we should make use of the flex layouts (*column, row, expanded, margin, etc). This is *0.3/0.4 for width, heights are not scalable enough for all screen sizes

lib/features/give/widgets/filter_suggestion_card.dart
```diff
@@ -8,70 +8,79 @@ class FilterSuggestionCard extends StatelessWidget {
...
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: [
+                  Container(
+                    margin: const EdgeInsets.symmetric(
+                      vertical: 5,
+                      horizontal: 5,
+                    ),
+                    child: Image.asset(
+                      isFocused ? activeIcon : icon,
+                      width: size.width * (isFocused ? 0.08 : 0.06),
```
MaikelStuivenberg, line 72: Also this is a bit ugly, I guess we can fix that differently and more scalable

alexbejann: Yep, it was existing. I gotta change it

MaikelStuivenberg (approved):

lib/features/give/widgets/choose_category.dart
alexbejann: Just tested on an iPhone SE and it's looks ok. But indeed I would have to change it in the end

PR #260, opened 2023-09-20
bug fix: KIDS 1359 Fix overflow issues

Daniela510:
## Description
Fix overflow issues in vpc flow

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -396,7 +396,13 @@ class APIService {
       '/givtservice/v1/PaymentProvider/checkoutsession/$guid/parent-control-validation',
     );
 
-    final response = await client.get(url);
+    final response = await client.get(
+      url,
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept': 'application/problem+json',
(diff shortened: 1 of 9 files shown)


lib/core/network/api_service.dart
```diff
@@ -396,7 +396,13 @@ class APIService {
       '/givtservice/v1/PaymentProvider/checkoutsession/$guid/parent-control-validation',
     );
 
-    final response = await client.get(url);
+    final response = await client.get(
```
MaikelStuivenberg, line 399: do we need to add that problem+json thing?

Daniela510: Endpoint was returning 500 "No MediaTypeFormatter is available to read an object of type 'Object' from content with media type 'application/problem+json'." so I thought this might fix it, but it works with out it now --- no, i guess we don't _need it_

pavelpavelko (approved):
lgtm)

PR #289, opened 2023-09-27
fix: bff 1041/recurring fixes

alexbejann:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/features/recurring_donations/create/widgets/create_recurring_donation_bottom_sheet.dart
+++ b/lib/features/recurring_donations/create/widgets/create_recurring_donation_bottom_sheet.dart
@@ -41,6 +41,7 @@ class CreateRecurringDonationBottomSheetView extends StatefulWidget {
 class _CreateRecurringDonationBottomSheetViewState
     extends State<CreateRecurringDonationBottomSheetView> {
   late TextEditingController amountController;
+  late TextEditingController turnsController;
 
   @override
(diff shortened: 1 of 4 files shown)


lib/features/recurring_donations/create/widgets/create_recurring_donation_bottom_sheet.dart
```diff
@@ -68,6 +72,12 @@ class _CreateRecurringDonationBottomSheetViewState
     final currencySymbol = NumberFormat.simpleCurrency(
       name: country.currency,
     ).currencySymbol;
+
+    if (user.country != Country.us.countryCode ||
```
pavelpavelko, line 76: just a thoughts: we have so many places with such or very similar code. idk how yet, but it would be great to incapsulate such logic at leas in Util class, but even there I think we could come up with some more generic solution

pavelpavelko (approved):
lgtm :)

lib/features/recurring_donations/create/widgets/create_recurring_donation_bottom_sheet.dart
alexbejann: Yes! I agree, let's make it as an improvement on the backlog

PR #299, opened 2023-09-29
fix: BFF-1331 Removed user presets from user ext and added into a separate…

alexbejann:
… class based on guid

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/features/amount_presets/models/user_presets.dart
+++ b/lib/features/amount_presets/models/user_presets.dart
@@ -1,8 +1,34 @@
 import 'package:equatable/equatable.dart';
 import 'package:givt_app/features/amount_presets/models/preset.dart';
 
+class AmountPresets extends Equatable {
+  const AmountPresets({
+    required this.presets,
+  });
+
+  factory AmountPresets.fromJson(Map<String, dynamic> json) => AmountPresets(
+        presets: (json['amountPresets'] as List<dynamic>)
+            .map((e) => UserPresets.fromJson(e as Map<String, dynamic>))
+            .toList(),
+      );
+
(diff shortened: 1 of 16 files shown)


lib/features/amount_presets/models/user_presets.dart
```diff
@@ -1,8 +1,34 @@
 import 'package:equatable/equatable.dart';
 import 'package:givt_app/features/amount_presets/models/preset.dart';
 
+class AmountPresets extends Equatable {
```
pavelpavelko, line 4: would be better to move it in a separate file amount_presets.dart?

lib/features/first_use/pages/welcome_page.dart
```diff
@@ -78,7 +78,8 @@ class _WelcomePageViewState extends State<WelcomePageView> {
               Padding(
                 padding: const EdgeInsets.symmetric(vertical: 15),
                 child: ElevatedButton(
-                  onPressed: () => Navigator.of(context).push(EmailSignupPage.route()),
+                  onPressed: () =>
+                      Navigator.of(context).push(EmailSignupPage.route()),
```
pavelpavelko, line 82: why native navigator instead of go_router approach is used here?

pavelpavelko (approved):
lgtm

lib/features/amount_presets/models/user_presets.dart
alexbejann: Yes, you're right!

PR #318, opened 2023-10-09
fix: BFF-1372 Added url check for country if it's available on the device

alexbejann:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/app/injection/injection.dart
+++ b/lib/app/injection/injection.dart
@@ -1,3 +1,4 @@
+import 'dart:convert';
 import 'dart:developer';
 import 'dart:io';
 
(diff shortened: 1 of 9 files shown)


lib/app/injection/injection.dart
```diff
@@ -53,6 +55,21 @@ Future<void> initAPIService() async {
...
+/// If there is, return the country of the user extension.
+/// If there is not, return the country of the device.
+Future<String> _checkCountry() async {
+  final prefs = await SharedPreferences.getInstance();
+
+  if (prefs.containsKey(UserExt.tag)) {
+    final userExtString = prefs.getString(UserExt.tag);
+    final user =
+        UserExt.fromJson(jsonDecode(userExtString!) as Map<String, dynamic>);
+    return user.country;
```
MaikelStuivenberg, line 68: Does this work for a temp user? And do we maybe want to check if country really is filled?

alexbejann: It works for a temp user, I tested it before

MaikelStuivenberg (approved):

PR #341, opened 2023-10-18
feat: kids 1052 edit childs profile

pavelpavelko:
## Description

- Fetch profile details endpoint is implemented; 
- Child details screen and logic is implemented; 
- Edit child screen is implemented (first draft); 
- App bar added for most of children screens; 
- Validation logic separated from create and edit screens; 
- Route utils replaced with Pages enum; 
- Other code improvements;
- Fetching name and date of birth via child details endpoint

## Type of Change
- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/app/routes/pages.dart
alexbejann, line 1: Love it!

lib/app/routes/routes.dart
alexbejann, line 1: 💯

lib/core/network/api_service.dart
```diff
@@ -454,6 +475,27 @@ class APIService {
...
+    if (response.statusCode >= 300) {
+      throw GivtServerFailure(
+        statusCode: response.statusCode,
+        body: jsonDecode(response.body) as Map<String, dynamic>,
+      );
+    }
+
+    final decodedBody = jsonDecode(response.body) as Map<String, dynamic>;
+    final itemMap = decodedBody['item'] as Map<String, dynamic>;
+    log(itemMap.toString());
```
alexbejann, line 495: Why do you log this here?

lib/features/children/create_child/mixins/child_date_of_birth_validator.dart
alexbejann, line 1: Maybe you can make use of the barrel files here as well. To make the imports nicer. Like the one from the routes 

`/create_child/mixins/mixins.dart` and add the exports of each file and then you import only once. If you need to implement another validator you can just add it into that file and keep the other imports clean 🥇 (as that would be adjusted by that single import)

lib/features/children/details/cubit/child_details_cubit.dart
```diff
@@ -0,0 +1,28 @@
...
+    emit(const ChildDetailsFetchingState());
+    try {
+      final response = await _childDetailsRepository.fetchChildDetails(profile);
+      emit(
+        ChildDetailsFetchedState(
+          profileDetails: response,
+        ),
+      );
+    } catch (error) {
+      emit(ChildDetailsErrorState(errorMessage: error.toString()));
```
alexbejann, line 25: maybe it's helpful to log this to kibana. But why don't you catch on the GivtServerFailure? as you have a body and statusCode there

lib/features/children/details/pages/child_details_page.dart
```diff
@@ -0,0 +1,95 @@
...
+
+class ChildDetailsPage extends StatelessWidget {
+  const ChildDetailsPage({
+    super.key,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final locals = AppLocalizations.of(context);
+    return BlocBuilder<ChildDetailsCubit, ChildDetailsState>(
```
alexbejann, line 20: I would try to not wrap the whole page into a builder but try to go as deep as possible

lib/features/children/details/pages/child_details_page.dart
```diff
@@ -0,0 +1,95 @@
...
+            actions: [
+              BackButton(
+                onPressed: () {
+                  context.pop();
+                  AnalyticsHelper.logEvent(
+                    eventName: AmplitudeEvents.backClicked,
+                  );
+                },
+              ),
+              const Spacer(),
```
alexbejann, line 33: Does this work on a smaller screen? Or isn't overflowing?

lib/features/children/details/pages/child_details_page.dart
```diff
@@ -0,0 +1,95 @@
...
+                    );
+                  },
+                ),
+            ],
+            automaticallyImplyLeading: false,
+            backgroundColor: state is ChildDetailsFetchedState
+                ? state.profileDetails.profile.monsterColor
+                : null,
+          ),
+          body: state is ChildDetailsFetchingState
```
alexbejann, line 57: I've seen a lot of these kind of conditionals into the registration part as well. Coming back to it after a while it becomes confusing and creates some overhead. What if you would put here the builder and then based on the state you return the desired widget?

lib/features/children/details/repositories/child_details_repository.dart
```diff
@@ -0,0 +1,18 @@
...
+mixin ChildDetailsRepository {
+  Future<ProfileExt> fetchChildDetails(Profile profile);
+}
+
+class ChildDetailsRepositoryImpl with ChildDetailsRepository {
+  ChildDetailsRepositoryImpl(this.apiService);
+  final APIService apiService;
+
+  @override
+  Future<ProfileExt> fetchChildDetails(Profile profile) async {
```
alexbejann, line 14: Why do you do the model operation in the repository? Which should only fetch and return the data? This breaks the single responsibility rule from SOLID principles. what do you think? I would do this in the cubit/bloc (as that should be the place for business operations)

lib/features/children/details/widgets/child_details_item.dart
```diff
@@ -0,0 +1,65 @@
...
+    required this.profileDetails,
+    super.key,
+  });
+
+  final ProfileExt profileDetails;
+
+  @override
+  Widget build(BuildContext context) {
+    final user = context.read<AuthCubit>().state.user;
+    final currency = NumberFormat.simpleCurrency(
```
alexbejann, line 21: Don't we have a Util method for this?

lib/features/children/details/widgets/child_details_item.dart
```diff
@@ -0,0 +1,65 @@
...
+      name: Util.getCurrencyName(
+        country: Country.fromCode(user.country),
+      ),
+    );
+
+    final size = MediaQuery.sizeOf(context);
```
alexbejann, line 27: haha Love it!

lib/features/children/details/widgets/child_giving_allowance_card.dart
```diff
@@ -0,0 +1,99 @@
...
+    this.onPressed,
+    super.key,
+  });
+
+  final ProfileExt profileDetails;
+  final void Function()? onPressed;
+
+  @override
+  Widget build(BuildContext context) {
+    final locals = AppLocalizations.of(context);
```
alexbejann, line 20: You can do `context.l10n`

lib/features/children/edit_child/cubit/edit_child_cubit.dart
```diff
@@ -0,0 +1,77 @@
...
+        EditChildInputState(
+          profileDetails: _profileDetails,
+          child: EditChild.fromProfileDetails(_profileDetails),
+        ),
+      );
+    });
+  }
+
+  final CreateChildRepository _createChildRepository;
+  final AppLocalizations _locals;
```
alexbejann, line 30: Oh, not sure if this is something we should do. Why do we need it here and not do it based on the state in the build method?

lib/features/children/edit_child/cubit/edit_child_cubit.dart
```diff
@@ -0,0 +1,77 @@
...
+  final ProfileExt _profileDetails;
+
+  bool _validateInput(EditChild child) {
+    final nameErrorMessage =
+        validateName(locals: _locals, name: child.firstName);
```
alexbejann, line 35: Oh.. yeah, don't think it's a good idea 👎

lib/features/children/edit_child/cubit/edit_child_cubit.dart
```diff
@@ -0,0 +1,77 @@
...
+    final allowanceErrorMessage =
+        validateGivingAllowance(locals: _locals, allowance: child.allowance);
```
alexbejann, line 37: Same

lib/features/children/edit_child/cubit/edit_child_cubit.dart
```diff
@@ -0,0 +1,77 @@
...
+        _profileDetails.profile.id,
+        child,
+      );
+      if (isChildUpdated) {
+        emit(
+          EditChildSuccessState(profileDetails: _profileDetails),
+        );
+      } else {
+        //TODO: POEditor
+        throw Exception('Cannot update child profile. Please try again later.');
```
alexbejann, line 69: Why don't you handle this with a state for the cubit? instead of throwing this exception

lib/features/children/edit_child/pages/edit_child_page.dart
```diff
@@ -0,0 +1,234 @@
...
+        '$currencySymbol${child.allowance.toStringAsFixed(0)}';
+    _allowanceController.text = allowanceText;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final locals = AppLocalizations.of(context);
+
+    final user = context.read<AuthCubit>().state.user;
+    final currency = NumberFormat.simpleCurrency(
```
alexbejann, line 70: Same, with the util method

lib/features/children/edit_child/pages/edit_child_page.dart
```diff
@@ -0,0 +1,234 @@
...
+      name: Util.getCurrencyName(country: Country.fromCode(user.country)),
+    );
+
+    return BlocConsumer<EditChildCubit, EditChildState>(
```
alexbejann, line 74: Same, you don't need this top level consumer and you can nest it into the body for instace and display from there the snackbars according to the state

lib/features/children/edit_child/pages/edit_child_page.dart
```diff
@@ -0,0 +1,234 @@
...
+          ),
+          body: state is EditChildUploadingState
+              ? const Center(
+                  child: CircularProgressIndicator(
+                    color: AppTheme.sliderIndicatorFilled,
+                  ),
+                )
+              : state is EditChildInputState ||
+                      state is EditChildInputErrorState
+                  ? SizedBox(
```
alexbejann, line 149: Yeah.... still not a big fan of this short hand conditionals. especially here you have a more complex one

lib/features/children/overview/models/profile.dart
```diff
@@ -63,4 +66,21 @@ class Profile extends Equatable {
...
+      monsterColor = AppTheme.childMonsterPurple;
+    } else if (pictureURL.contains('monster3')) {
+      monsterColor = AppTheme.childMonsterBlue;
+    } else if (pictureURL.contains('monster4')) {
+      monsterColor = AppTheme.childMonsterOrange;
+    } else {
+      monsterColor = Colors.transparent;
+    }
+    return monsterColor.withAlpha((255 * 0.1).toInt());
+  }
```
alexbejann, line 85: Ih, that's indeed a workaround

lib/features/children/overview/widgets/child_item.dart
```diff
@@ -1,130 +1,145 @@
...
-    required this.pending,
+    required this.profile,
     super.key,
   });
 
-  final String name;
-  final double total;
-  final double pending;
-  final String currencySymbol;
+  final Profile profile;
```
alexbejann, line 18: Do you really need the whole object? I see you're only using the pending

lib/features/children/overview/widgets/child_item.dart
```diff
@@ -1,130 +1,145 @@
...
 
   @override
   Widget build(BuildContext context) {
     final locals = AppLocalizations.of(context);
-    return Card(
-      shape: RoundedRectangleBorder(
-        borderRadius: BorderRadius.circular(15),
+
+    final user = context.read<AuthCubit>().state.user;
+    final currency = NumberFormat.simpleCurrency(
```
alexbejann, line 25: Util

lib/features/children/create_child/mixins/child_date_of_birth_validator.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:givt_app/l10n/l10n.dart';
+
+mixin ChildDateOfBirthValidator {
+  String? validateDateOfBirth({
+    required AppLocalizations locals,
```
alexbejann, line 5: Not sure if this is going to be a good way of doing the localization as you have to pass it through a lot of objects

alexbejann (changes requested):
I believe there could be some improvements. But that's always the case. I guess the most crucial ones are those conditionals in some pages

lib/core/network/api_service.dart
pavelpavelko: forgot initially, removed now, thanks

lib/features/children/details/cubit/child_details_cubit.dart
pavelpavelko: yep, Loggit added and switched to GivtServerFailure instead of just error

lib/features/children/details/pages/child_details_page.dart
pavelpavelko: goop point in terms of optimisation, however I need builder on that level since I use state for scaffold's app bar.

lib/features/children/details/pages/child_details_page.dart
pavelpavelko: Do you see any issues with Spacer here? not sure how it can lead to overflow.. The layout itself also quite flexible. Not sure where is the issue here?

lib/features/children/details/pages/child_details_page.dart
pavelpavelko: hmm, questionable for me. I like this style in general

lib/features/children/details/repositories/child_details_repository.dart
pavelpavelko: afaik repository can and should convert raw data (from provider) to model and then return it to cubit. Or I missed something?

lib/features/children/details/widgets/child_details_item.dart
pavelpavelko: good point, I've extended the util class a bit and replaced code above with it in most of the places, thanks

lib/features/children/details/widgets/child_giving_allowance_card.dart
pavelpavelko: great, that approach looks much more clear, thank you

lib/features/children/edit_child/cubit/edit_child_cubit.dart
pavelpavelko: I use it for getting errors from validators

lib/features/children/edit_child/cubit/edit_child_cubit.dart
pavelpavelko: emm, what exactly you don't like? :) could you please elaborate and maybe provide some alternatives? :)

lib/features/children/edit_child/cubit/edit_child_cubit.dart
pavelpavelko: please see my question above

lib/features/children/edit_child/cubit/edit_child_cubit.dart
pavelpavelko: you are right, addressed

lib/features/children/edit_child/pages/edit_child_page.dart
pavelpavelko: same here, I need a state object on an AppBar level

lib/features/children/overview/widgets/child_item.dart
pavelpavelko: I like such approach. Do you see any issues here?

lib/features/children/overview/widgets/child_item.dart
pavelpavelko: fixed

alexbejann (approved):

lib/features/children/create_child/mixins/child_date_of_birth_validator.dart
pavelpavelko: https://nfcollect.atlassian.net/browse/KIDS-1514

PR #344, opened 2023-10-18
Infinite loading on child details page is fixed

pavelpavelko:
## Description

Infinite loading on child details page is fixed

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/features/children/details/cubit/child_details_cubit.dart
+++ b/lib/features/children/details/cubit/child_details_cubit.dart
@@ -1,6 +1,5 @@
 import 'package:bloc/bloc.dart';
 import 'package:equatable/equatable.dart';
-import 'package:givt_app/core/failures/failures.dart';
 import 'package:givt_app/core/logging/logging.dart';
 import 'package:givt_app/features/children/details/models/profile_ext.dart';
 import 'package:givt_app/features/children/details/repositories/child_details_repository.dart';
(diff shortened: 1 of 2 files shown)


lib/features/children/details/cubit/child_details_cubit.dart
```diff
@@ -25,7 +25,8 @@ class ChildDetailsCubit extends Cubit<ChildDetailsState> {
       );
     } on GivtServerFailure catch (error) {
       await LoggingInfo.instance.error(error.toString());
-
+      emit(ChildDetailsErrorState(errorMessage: error.toString()));
```
MaikelStuivenberg, line 28: There is not difference between the first and second catch (except for logging) so I would combine it :)

pavelpavelko: the first catches BE-related (responseCode >= 300) issues, so I think it's better to logging it, however we could have other errors (like we had yesterday in the evening) that we need to handle on presentation level as well, but not sure about logging since it's something that we need to catch thile development/testing. I seems to me similar to checked vs unchecked exceptions in java. However I'm open to your suggestion here, what would you suggest? :)

pavelpavelko: working on another feature with similar try catch I realise that it would be better from UX perspective to come back to the previous generalised approach with catching all errors instead of just GivtServerFailure

pavelpavelko: pushed, please take a look)

MaikelStuivenberg (approved):

PR #350, opened 2023-10-20
fix: BFF-1393 Display multiple possible actions on the dialog when a inactive qr-code is scanned 

alexbejann:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/features/give/bloc/give/give_bloc.dart
+++ b/lib/features/give/bloc/give/give_bloc.dart
@@ -74,6 +74,10 @@ class GiveBloc extends Bloc<GiveEvent, GiveState> {
 
       await _checkQRCode(mediumId: mediumId, emit: emit);
 
+      if (state.status == GiveStatus.beaconNotActive) {
+        return;
+      }
+
       await _processGivts(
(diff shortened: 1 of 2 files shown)


lib/features/give/bloc/give/give_bloc.dart
```diff
@@ -554,6 +562,10 @@ class GiveBloc extends Bloc<GiveEvent, GiveState> {
     if (mediumId.split('.').last.contains('b6')) {
       return;
     }
+    if (mediumId.split('.').last.startsWith('6')) {
```
MaikelStuivenberg, line 565: Why?

MaikelStuivenberg (approved):

lib/features/give/bloc/give/give_bloc.dart
alexbejann: Check slack. I message you the reason. Also, i feel like it's an implementation bottleneck that we would have to deal with in the future as they have a designated instance type in the medium id

PR #365, opened 2023-10-25
Pin version numbers in pipeline

jul-reb:


.github/workflows/ios.yml
```diff
@@ -50,7 +57,7 @@ jobs:
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1.118.0
         with: 
-          ruby-version: '3.0'
+          ruby-version: ${{ inputs.ruby-version }}
```
alexbejann, line 60: Close the other PR related to the ruby version number

alexbejann (approved):

.github/workflows/ios.yml
jul-reb: if it's my rbenv PR I'd rather not :)
That's something else that I'll explain friday in the demo.

PR #373, opened 2023-10-26
Feature: kids-1456 child donation approval

pavelpavelko:
## Description

kids-1456 child donation approval is implemented

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/features/children/family_history/widgets/pending_donation_widget.dart
```diff
@@ -1,85 +1,129 @@
...
+                        maxLines: 2,
+                        overflow: TextOverflow.ellipsis,
+                        softWrap: true,
+                        style:
+                            Theme.of(context).textTheme.titleMedium?.copyWith(
+                                  color: AppTheme.childHistoryPendingDark,
+                                  fontFamily: 'Raleway',
+                                  fontWeight: FontWeight.bold,
+                                  fontSize: 18,
+                                ),
```
alexbejann, line 100: Wouldn't be better to have this in a separate widget? if you keep using that fontFamily in more places?

lib/features/children/parental_approval/dialogs/parental_approval_dialog.dart
```diff
@@ -0,0 +1,82 @@
...
+    final size = MediaQuery.sizeOf(context);
+    return BlocProvider(
+      create: (_) => ParentalApprovalCubit(
+        donation: donation,
+        decisionRepository: getIt(),
+      ),
+      child: Center(
+        child: SizedBox(
+          width: size.width * 0.75,
+          height: size.width * 0.75,
```
alexbejann, line 32: Why based on width and not on height?

lib/features/children/parental_approval/widgets/parental_approval_declined_page.dart
```diff
@@ -0,0 +1,50 @@
...
+  final ChildDonation donation;
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [
+        Text(
+          '${context.l10n.childParentalApprovalDeclinedTitlePrefix}${donation.name}${context.l10n.childParentalApprovalDeclinedTitlePostfix}',
```
alexbejann, line 22: If you want to compose strings like this it's better to do it in PO editor. By adding "My string {value0} with second {value1}" and in code you just call "context.l10n" and then you get the option to add 2 parameters and in this way if the text changes you don't have to make code changes.

alexbejann: [Example](https://github.com/givtnl/givt-app/blob/528c626282592cf10b82f5e038967f41b41e27d0/lib/l10n/arb/app_en_US.arb#L130C18-L130C18)

Like this one for example

alexbejann (approved):
I just left some questions and suggestions if you want to address them. I would address the one where you compose the strings into the code

lib/features/children/family_history/widgets/pending_donation_widget.dart
```diff
@@ -1,85 +1,129 @@
...
+          },
+        );
+        if (mounted) {
+          final decisionMade = await showDialog<bool>(
+            context: context,
+            builder: (context) => ParentalApprovalDialog(
+              donation: widget.donation,
+            ),
+          );
+          if (true == decisionMade && mounted) {
```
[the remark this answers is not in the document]
pavelpavelko: decisionMade is nullable, that's why I'm checking it by comparing with true :)

lib/features/children/parental_approval/dialogs/parental_approval_dialog.dart
pavelpavelko: yep, I just need square dialog with horizontal margins, so I calculate its size by screen width

lib/features/children/parental_approval/widgets/parental_approval_declined_page.dart
pavelpavelko: super nice, didn't know about that, thank you!

pavelpavelko: addressed. once again thanks, now it looks much nicer :)

PR #385, opened 2023-10-29
Improve Github Actions (BFF-1423) 

MaikelStuivenberg:
## Description
Automatic assign labels, template for release notes + optimising versioning etc :)


.github/workflows/android.yml
```diff
@@ -52,9 +49,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Ruby
-        uses: ruby/setup-ruby@v1.118.0
+        uses: ruby/setup-ruby@v1
         with: 
-          ruby-version: ${{ inputs.ruby-version }}
```
MaikelStuivenberg, line 57: this is not needed as we have a .ruby-version file

.github/workflows/android.yml
```diff
@@ -52,9 +49,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Ruby
-        uses: ruby/setup-ruby@v1.118.0
+        uses: ruby/setup-ruby@v1
```
MaikelStuivenberg, line 52: only v1 is recommended by the plugin

.github/workflows/main.yaml
```diff
@@ -38,8 +38,7 @@ jobs:
       MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
       LOGIT_API_KEY: ${{ secrets.LOGIT_API_KEY }}
     with:
-      flutter-version: '3.13.7'
-      ruby-version: '2.6.0'
+      flutter-version: '3.13.x'
```
MaikelStuivenberg, line 41: In this way we allow bugfixes

alexbejann (approved):

.github/workflows/android.yml
jul-reb: That is not how it works sadly!
The default ruby installer in github is not aware of the .ruby-version file.

MaikelStuivenberg: According to the docs it is

jul-reb: nvm!

PR #395, opened 2023-11-01
BFF-645 Fixed issue where the issue didn't give previous year

alexbejann:
## Description

We should avoid making calculations for previous year if the user is new or didn't donate to previous year

### DIFF
--- a/lib/features/personal_summary/yearly_overview/cubit/yearly_overview_state.dart
+++ b/lib/features/personal_summary/yearly_overview/cubit/yearly_overview_state.dart
@@ -116,8 +116,7 @@ final class YearlyOverviewState extends Equatable {
       );
     }
 
-    final copyOfMergedList = mergedList;
-    for (var it in copyOfMergedList) {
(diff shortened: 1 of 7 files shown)


lib/features/personal_summary/yearly_overview/cubit/yearly_overview_state.dart
```diff
@@ -113,7 +116,8 @@ final class YearlyOverviewState extends Equatable {
       );
     }
 
-    for (var it in mergedList) {
+    final copyOfMergedList = mergedList;
```
MaikelStuivenberg, line 119: If I'm correct, this doesn't copy a list, but will make a reference to the other list.

If you want to copy I think you need to do:
Copylist = [...mergedlist]

And even that is not fully a copy. It's only creating a new list with the same objects (references to those objects)

So depending on what you need you can keep it or you need to change it :)

MaikelStuivenberg (approved):

PR #403, opened 2023-11-08
Beacon improvements (BFF-1454)

MaikelStuivenberg:
## Description
restart scan process after timeout + make sure that we stop processing scan results on manual flow

### DIFF
--- a/lib/features/give/pages/bt_scan_page.dart
+++ b/lib/features/give/pages/bt_scan_page.dart
@@ -20,8 +20,8 @@ class BTScanPage extends StatefulWidget {
 }
 
 class _BTScanPageState extends State<BTScanPage> {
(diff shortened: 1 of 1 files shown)


lib/features/give/pages/bt_scan_page.dart
```diff
@@ -22,13 +22,26 @@ class BTScanPage extends StatefulWidget {
 class _BTScanPageState extends State<BTScanPage> {
   bool _isVisible = false;
 
+  // We need to keep track of the scanning state to prevent scans when the user is going into the manual flow
+  bool _isSearching = false;
```
alexbejann, line 26: The whole class is already private, is it needed to have these ones private as well? 🤔

alexbejann (approved):

PR #405, opened 2023-11-08
PushnotificationId is not send to our db when user account is created (BFF-1456)

MaikelStuivenberg:
## Description
Set notificationid when temp account is transformed to registered


lib/shared/models/user_ext.dart
```diff
@@ -281,6 +282,7 @@ class UserExt extends Equatable {
         personalInfoRegistered,
         pinSet,
         multipleCollectsAccepted,
+        notificationId,
```
alexbejann, line 285: Good one, didn't notice i forgot to add it to the props 😅

alexbejann (approved):

PR #419, opened 2023-11-15
BFF-1459 Navigate to the the payload screen

alexbejann:
## Description
When you receive a notification you should be redirected to the screen accordingly. Therefore, I am uisng the routeDelegate to do so.

### DIFF
--- a/lib/core/notification/notification_service.dart
+++ b/lib/core/notification/notification_service.dart
@@ -21,6 +21,7 @@ Future<void> _navigateToScreen(NotificationResponse details) async {
     case 'RecurringDonation':
       await LoggingInfo.instance
           .info('Navigating to recurring donation screen');
+      AppRouter.router.goNamed(Pages.recurringDonations.name);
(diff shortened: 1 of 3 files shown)


lib/core/notification/notification_service.dart
```diff
@@ -4,12 +4,34 @@ import 'package:firebase_messaging/firebase_messaging.dart';
...
 import 'package:timezone/timezone.dart' as tz;
 
+/// Navigate to the screen based on the notification payload
+/// This is called when the user taps on the notification
+@pragma('vm:entry-point')
+Future<void> _navigateToScreen(NotificationResponse details) async {
+  final payload = details.payload;
+  switch (payload) {
+    case 'RecurringDonation':
+      await LoggingInfo.instance
```
MaikelStuivenberg, line 22: Maybe its good to implement them or leave the logging out. 
Right now we would get logging in Logit, but the action will never happen

MaikelStuivenberg (approved):

PR #425, opened 2023-11-21
Add amplitude tracking for Personal Summary 

MaikelStuivenberg:


lib/features/auth/repositories/auth_repository.dart
```diff
@@ -107,6 +108,14 @@ class AuthRepositoyImpl with AuthRepositoy {
         UserExt.tag,
         jsonEncode(newUserExt.toJson()),
       );
+
```
pavelpavelko, line 111: this code looks like duplicate of the ` Future<UserExt> fetchUserExtension(String guid) async {` below.
Can we reuse it in this place?

pavelpavelko (approved):

PR #436, opened 2023-11-28
BFF-1471 Open app from external link to a certain page using the path

alexbejann:
## Description

By using the path from the enum "Pages" we can open the app using an external link as the one provided below
https://givtapp.net/download?page=personal-summary

The implementation is meant to use the query parameter "page" to navigate to a named route dynamically


MaikelStuivenberg (approved):
@alexbejann can you add amplitude logging as well? :)

PR #446, opened 2023-12-06
Bug: KIDS-110-add safe area to VPC flow

pavelpavelko:
## Description
- VPC flow pages are slightly refactored; 
- SafeArea added to the all VPC pages; 
- SafeArea added to the Create child screen; 
- Snackbar helper added;


Daniela510 (approved):
Ran in emulator to see, all looks good!

PR #463, opened 2023-12-14
BFF-1485 Do not redirect the user to the home page if he's already there

alexbejann:

### DIFF
--- a/lib/utils/auth_utils.dart
+++ b/lib/utils/auth_utils.dart
@@ -38,12 +38,7 @@ class AuthUtils {
       }
       await context.read<AuthCubit>().refreshSession();
 
-      /// Because the session is refreshed and the router needs to respond
-      /// we will delay the navigation by 100 milliseconds.
-      Future.delayed(
-        const Duration(milliseconds: 100),
-        navigate,
-      );
+      navigate();
     } on PlatformException catch (e) {
       await LoggingInfo.instance.info(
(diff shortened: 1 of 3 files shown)


lib/utils/auth_utils.dart
```diff
@@ -37,7 +37,13 @@ class AuthUtils {
         return;
       }
       await context.read<AuthCubit>().refreshSession();
-      navigate();
+
+      /// Because the session is refreshed and the router needs to respond
+      /// we will delay the navigation by 100 milliseconds.
+      Future.delayed(
```
MaikelStuivenberg, line 43: This doesnt sound like a really good solution. It probably works but we don't know how long exactly we need to wait. It might even be different per phone (slow/older phones could be slower?)

MaikelStuivenberg (changes requested):

lib/utils/auth_utils.dart
MaikelStuivenberg: Is there a way to do this differently?

alexbejann: Nope, it's not ideal. I am thinking of a different way to do it. As this is cause from the routing as we have the refresh part

MaikelStuivenberg (approved):

lib/utils/auth_utils.dart
alexbejann: Should be fine now

PR #470, opened 2023-12-17
BFF-1053 Giving confirm page advertisements

alexbejann:
## Description

WIP
1. Added advertisement repository
2. Added Models
3. Added api call


TammiLion:

PR #485, opened 2023-12-23
BFF-659 Added The app From A-Z and Share button

alexbejann:
## Description

1. Added From A-Z menu item
2. Added Share Givt button menu item


TammiLion:

PR #495, opened 2023-12-28
BFF-1498 Adjusted dimensions and cleaned the recurring donation page

alexbejann:
## Description

1. Adjusted dimensions of the recurring donation list item
2. Removed redundant parameters
3. Fixed overflow of the text on smaller screens


TammiLion:

PR #496, opened 2023-12-28
Feature/kids 178 update registration flow

Daniela510:
## Description

- Change last name and surname string depending on flow
- Change registration dialog text based on country flow not language
- Add waiting text while stripe is loading


lib/features/registration/pages/signup_page.dart
```diff
@@ -243,7 +250,9 @@ class _SignUpPageState extends State<SignUpPage> {
               style:
                   Theme.of(context).textTheme.bodyLarge?.copyWith(fontSize: 16),
               decoration: InputDecoration(
-                hintText: AppLocalizations.of(context).lastName,
+                hintText: isUS
```
MaikelStuivenberg, line 253: Is surname/lastname not the same thing? Do we need to make a difference here based on country?

lib/features/give/pages/home_page.dart
```diff
@@ -153,7 +153,8 @@ class _HomePageState extends State<HomePage> {
                 if (!auth.user.needRegistration || auth.user.mandateSigned) {
                   return;
                 }
-                _buildNeedsRegistrationDialog(context);
+                final isUS = auth.user.country == Country.us.countryCode;
```
MaikelStuivenberg, line 156: its not needed to pass this information. User object is also available in the _build... method

lib/features/registration/pages/signup_page.dart
Daniela510: surename is UK english, lastname is US. In the slack thread we discussed changes should be based on selected country not on phone language.
https://nfcollect.slack.com/archives/C04PS1D7ARL/p1703596032962579

lib/features/give/pages/home_page.dart
Daniela510: Improved

lib/features/registration/pages/signup_page.dart
MaikelStuivenberg: If lastname is US english we should change this based on phone language (Users that have US as language on their phone always want to see that, even if they use Givt in an other country). 

For the changes that are showing givt4kids specific things its indeed better to do it based on country :) as we only want to have that flow if its a US Givt user.

Daniela510: okay, the term is "surname" and will be surname in UK& default english, but "Last name" if the phone language is set to US eng.

lib/features/give/pages/home_page.dart
```diff
@@ -187,15 +187,19 @@ class _HomePageState extends State<HomePage> {
     BuildContext context,
   ) {
     final user = context.read<AuthCubit>().state.user;
+    final isUS = user.country == Country.us.countryCode;
```
alexbejann, line 190: I was thinking to move this thing into the the Country enum extension as this is soo common in the entire codebase. So I am thinking of something like "Country.isUS" and you can just do Countr.fromCode(user.countryCode). And use the "country.isUS" whenever needed. What do you think?

lib/features/registration/pages/signup_page.dart
```diff
@@ -137,7 +137,14 @@ class _SignUpPageState extends State<SignUpPage> {
           builder: (context, state) {
             return isLoading
                 ? const Center(
-                    child: CircularProgressIndicator(),
+                    child: Column(
+                      mainAxisAlignment: MainAxisAlignment.center,
+                      children: [
+                        Text('Hold on, we are creating your account...'),
+                        SizedBox(height: 16),
+                        CircularProgressIndicator(),
```
alexbejann, line 145: I started making it adaptive whenever I encounter this "CircularProgressIndicator.adaptive()"

lib/features/registration/pages/signup_page.dart
```diff
@@ -137,7 +137,14 @@ class _SignUpPageState extends State<SignUpPage> {
...
+                      ],
+                    ),
                   )
                 : _buildSignUpForm(locals, size, isUs);
```
alexbejann, line 149: Also, these things I started putting them into separate widgets as it helps in terms in performance and also is not going to be re-rendered by the Engine

lib/features/registration/pages/signup_page.dart
Daniela510: What does .adaptive() do? It just visually looks different?

lib/features/give/pages/home_page.dart
Daniela510: Like this? https://github.com/givtnl/givt-app/pull/498

MaikelStuivenberg (approved):

lib/features/registration/pages/signup_page.dart
alexbejann: It's adapting the progress indicator to the platform. If iOS it display the cupertino progress indicator for Android is the Material one. We would have to do the same for dialogs as we use atm the cupertino dialog

lib/features/give/pages/home_page.dart
alexbejann: @Daniela510 There's a method in the Country enum you can just do "Country.fromCode(user.country).isUS"

PR #503, opened 2024-01-01
Feature/kids 234 and kids 231 merge vpc and child creation

Daniela510:
## Description

- Remove VPC as separate feature, it is now integrated into the add-member endpoint.
- Create the new child creation form, vpc page & success page


lib/features/children/add_member/cubit/add_member_cubit.dart
```diff
@@ -0,0 +1,54 @@
...
+  void goToVPC({required Child child}) {
+    emit(ConfirmVPCState(child: child));
+  }
+
+  void goToInput({required Child child}) {
+    emit(AddMemberInputState(child: child));
+  }
+
+  Future<void> createChildWithVPC() async {
+    if (state.child != null) {
```
alexbejann, line 27: You can do an early return and emit the else statement. Avoiding the unnecessary nesting.
```
if (state.child == null) {
emit(
  const AddMemberExternalErrorState(errorMessage: 'error'),
);
return;
}
```

lib/features/children/add_member/cubit/add_member_state.dart
```diff
@@ -0,0 +1,40 @@
...
+sealed class AddMemberState extends Equatable {
+  const AddMemberState({
+    this.child,
+  });
+
+  final Child? child;
+
+  @override
+  List<Object?> get props => [child];
+}
```
alexbejann, line 12: Wouldn't be easier to make a single state for this cubit? instead of just passing around the "child"?

lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -0,0 +1,346 @@
...
+      },
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final formKey = GlobalKey<FormState>();
+    final user = context.read<AuthCubit>().state.user;
+    final currency = NumberFormat.simpleCurrency(
+      name: Util.getCurrencyName(country: Country.fromCode(user.country)),
```
alexbejann, line 90: You have the Currency Name in the 
```Country.fromCode(user.country).currency```

File: lib/features/children/add_member/pages/member_main_scaffold_page.dart [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+Widget _createPage(AddMemberState state, BuildContext context) {
+  if (state is ConfirmVPCState) {
+    return const VPCPage();
+  } else if (state is AddMemberInputState || state is AddMemberInitial) {
+    return const CreateMemberPage();
+  } else if (state is AddMemberSuccessState) {
+    return const AddMemeberSuccessPage();
+  } else {
+    return Container();
+  }
```
alexbejann, line 63: A switch would be better in this case by using the `state.runtimeType` and then check the state

pubspec.yaml
```diff
@@ -55,6 +55,7 @@ dependencies:
   shared_preferences: ^2.2.0
   sprintf: ^7.0.0
   sticky_headers: ^0.3.0+2
+  store_redirect: ^2.0.1
```
alexbejann, line 58: I am pretty sure this package is redundant and at the same time is unmaintained. You can achieve the same thing with the url_launcher

lib/features/children/add_member/pages/success_add_member_page.dart
```diff
@@ -0,0 +1,112 @@
...
+          Column(
+            children: [
+              Padding(
+                padding: const EdgeInsets.only(left: 15, right: 15),
+                child: ElevatedButton(
+                  onPressed: () {
+                    StoreRedirect.redirect(
+                      androidAppId: 'net.givtapp.kids',
+                      iOSAppId: '1658797002',
+                    );
```
alexbejann, line 65: I left a comment for this package

lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -0,0 +1,346 @@
...
+      return;
+    }
+    setState(() {
+      _allowanceController--;
+    });
+  }
+
+  @override
+  void initState() {
+    final memeberCubit = context.read<AddMemberCubit>();
```
alexbejann, line 52: Is this a typo?

lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -0,0 +1,346 @@
...
+                  )
+                : BorderRadius.only(
+                    topRight: Radius.circular(4),
+                    bottomRight: Radius.circular(4),
+                  ),
+            border: isSelected
+                ? Border.all()
+                : Border.fromBorderSide(
+                    BorderSide(
+                      color: Colors.grey.shade300,
```
alexbejann, line 173: Don't we have a custom color for this grey? We different colors for grey, it would be nice to centrelize this

lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -0,0 +1,346 @@
...
+              if (formKey.currentState!.validate()) {
+                _createChildProfile();
+              }
+            }
+          : () {},
+      style: ElevatedButton.styleFrom(
+        backgroundColor: enabled ? AppTheme.givtLightGreen : Colors.grey,
+      ),
+      child: Text(
+        "Continue",
```
alexbejann, line 338: I noticed that the ```'``` and the ```"``` are not consistent

lib/features/children/add_member/cubit/add_member_state.dart
Daniela510: oh yeah, with the status! Thanks!

lib/features/children/add_member/pages/add_member_form_page.dart
Daniela510: Yes, im not paying much attention to it since i still have to add all the strings to POEditor

lib/features/children/add_member/pages/add_member_form_page.dart
Daniela510: `Colors.grey.shade300` is used in the AppTheme to define the border colors (including for the text fields). The textformfield border has to match the hint/label text color, so will update it there instead.

alexbejann (approved):

PR #509, opened 2024-01-05
Refactor category selection cards

MaikelStuivenberg:
## Description
I splitted up the card as they a different in multiple ways (text didn't fit anymore and wasn't always aligned everywhere)


pavelpavelko (approved):
lgtm :)

PR #512, opened 2024-01-05
Fix: no children redirection

Daniela510:


lib/features/children/add_member/widgets/add_member_form.dart
```diff
@@ -360,7 +360,7 @@ class _AddMemberFormState extends State<AddMemberForm> {
             },
             controller: _ageController,
             hintText: context.l10n.ageKey,
-            keyboardType: TextInputType.number,
+            keyboardType: const TextInputType.numberWithOptions(signed: true),
```
MaikelStuivenberg, line 363: As I already said on Slack, this solution isn't gonna work. This will bring up a normal keyboard

MaikelStuivenberg (changes requested):

PR #515, opened 2024-01-08
Improvement: Kids-239 Replace Icon button with Gesture Detector

Daniela510:
## Description

Add timer that quickly increases/decreases the _allowanceCounter when button is held down

### DIFF
--- a/lib/features/children/add_member/widgets/add_member_form.dart
+++ b/lib/features/children/add_member/widgets/add_member_form.dart
@@ -37,10 +37,17 @@ class _AddMemberFormState extends State<AddMemberForm> {
   final formKeyParent = GlobalKey<FormState>();
   Timer? _timer;
   Duration _heldDuration = Duration.zero;
+  int tapTime = 240;
+  int holdDownDuration = 1000;
+  int holdDownDuration2 = 2000;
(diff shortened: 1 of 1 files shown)


lib/features/children/add_member/widgets/add_member_form.dart
```diff
@@ -34,22 +35,69 @@ class _AddMemberFormState extends State<AddMemberForm> {
...
+  @override
+  void dispose() {
+    _stopTimer();
+    super.dispose();
+  }
 
   void _incrementCounter() {
-    if (_allowanceController > 998) {
+    final isHeldDurationShortEnoughForIncrement1 =
+        (_heldDuration.inMilliseconds > 1000) && _allowanceController > 994;
```
MaikelStuivenberg, line 61: Maybe it would be better to put all the numbers into constants / variables. 
Its really easy to make mistake right now if you wanna change one of the values.

Also the 994, 984, 998 feels like it should be a calculation instead of hard values. (if I understand it correctly)

MaikelStuivenberg (approved):

MaikelStuivenberg (commented):
I approved it instead putting in comments

PR #525, opened 2024-01-10
Improvement/kids 249 empty history and just parents

Daniela510:
## Description

If a user added only parents or if the children's first allowance did not get deducted yet: show empty history state.


lib/features/children/family_history/family_history.dart
```diff
@@ -78,6 +79,26 @@ class FamilyHistory extends StatelessWidget {
                       return getDivider(state, index);
                     },
                   ),
+                  if (state.history.isEmpty)
```
pavelpavelko, line 82: I would suggest to extract it to a separate HistoryEmptyStateWidget (name is just en example)

lib/features/children/family_history/family_history.dart
```diff
@@ -78,6 +79,26 @@ class FamilyHistory extends StatelessWidget {
...
+                        'assets/images/empty_lines.svg',
+                        width: size.width * 0.95,
+                      ),
+                    ),
+                  if (state.history.isEmpty)
+                    Center(
+                      child: Text(
+                        context.l10n.emptyChildrenDonations,
+                        textAlign: TextAlign.center,
+                        style: const TextStyle(
```
pavelpavelko, line 94: shall we use predefined TextStyles instead?

lib/features/children/overview/pages/family_overview_page.dart
```diff
@@ -37,7 +37,8 @@ class FamilyOverviewPage extends StatelessWidget {
           appBar: AppBar(
             centerTitle: false,
             title: state is FamilyOverviewUpdatedState &&
-                    state.profiles.where((p) => p.type == 'Child').isEmpty
+                    state.profiles.where((p) => p.type == 'Child').isEmpty &&
```
pavelpavelko, line 40: can we incapsulate this logic somewhere in FamilyOverviewUpdatedState class with simple and readable getters? like:
`  List<Profile> get children {
    return profiles.where((p) => p.type == 'Child').toList();
  }

  List<Profile> get adults {
    return profiles.where((p) => p.type == 'Parent').toList();
  }

  bool get hasChildren {
    return children.isEmpty;
  }

  bool get isAdultSingle {
    return adults.length == 1;
  }
`
Also, taking into account further scale on grandparents or even more, shall we use more generic type name like Adult instead of Parent?

lib/features/children/overview/pages/family_overview_page.dart
```diff
@@ -86,23 +89,39 @@ class FamilyOverviewPage extends StatelessWidget {
...
 
+  Widget buildFamilyOverviewBody(
+      FamilyOverviewState state, BuildContext context) {
+    if (state is FamilyOverviewLoadingState) {
+      return const ChildrenLoadingPage();
+    }
+
+    if (state is FamilyOverviewUpdatedState) {
+      final hasNoChildren =
+          state.profiles.where((p) => p.type == 'Child').isEmpty;
```
pavelpavelko, line 107: we can use getters proposed above here as well

lib/features/children/overview/pages/family_overview_page.dart
Daniela510: 'Parent'/'Child' matches the BE type, I agree Adult would be better

Daniela510: bool get hasChildren {
return children.is**Not**Empty;
}

lib/features/children/family_history/widgets/empty_history_stack.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_svg/svg.dart';
+import 'package:givt_app/l10n/l10n.dart';
+
+class EmptyHistoryWidget extends StatelessWidget {
```
pavelpavelko, line 5: the name of the file is different then the class name

lib/features/children/family_history/widgets/empty_history_stack.dart
```diff
@@ -0,0 +1,34 @@
...
+            'assets/images/empty_lines.svg',
+            width: size.width * 0.95,
+          ),
+        ),
+        Center(
+          child: Text(
+            context.l10n.emptyChildrenDonations,
+            textAlign: TextAlign.center,
+            style: Theme.of(context).textTheme.bodyMedium?.copyWith(
+                  fontSize: 14,
```
pavelpavelko, line 25: can we change to bodySmall (or other?) instead changing fontSize directly?

pavelpavelko (approved):
looks good with few small things)

PR #526, opened 2024-01-10
Kids-254 Continue without VPC if not creating any children

Daniela510:
## Description

Create member without vpc if no children are being created.


lib/features/children/add_member/cubit/add_member_cubit.dart
```diff
@@ -49,6 +49,16 @@ class AddMemberCubit extends Cubit<AddMemberState> {
   }
 
   void allFormsFilled() {
+    if (state.members.where((element) => element.type == 'Child').isEmpty) {
```
pavelpavelko, line 52: another good candidate for getters proposed in another ;)

pavelpavelko: also I'm a bit confused that now we have both Profile and Member models in the app.. Could you please provide more historical context here?

pavelpavelko (approved):
looks good with one small suggestion and also more general question (but it should not block merge oc)

lib/features/children/add_member/cubit/add_member_cubit.dart
Daniela510: Member is the model we need to add a member to the family group: children or adults. Profile is what we get returned from BE including monster picture & wallet amount.

PR #528, opened 2024-01-10
Improvement: kids-255 add text to loading state

Daniela510:

### DIFF
--- a/lib/features/children/add_member/pages/add_member_form_page.dart
+++ b/lib/features/children/add_member/pages/add_member_form_page.dart
@@ -131,7 +131,6 @@ class _CreateMemberPageState extends State<CreateMemberPage> {
       },
       builder: (context, state) {
         if (state.status == AddMemberStateStatus.loading) {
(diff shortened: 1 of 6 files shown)


lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -131,7 +131,20 @@ class _CreateMemberPageState extends State<CreateMemberPage> {
       },
       builder: (context, state) {
         if (state.status == AddMemberStateStatus.loading) {
-          return const Center(child: CircularProgressIndicator());
+          //here
```
pavelpavelko, line 134: do we need this comment?))

pavelpavelko (approved):
lgtm)

lib/features/children/add_member/pages/add_member_form_page.dart
Daniela510: No i forgot to delete

PR #535, opened 2024-01-11
Improvement: kids-258 registration success screen

Daniela510:
## Description

- Create a success screen instead of a dialog
- Create check animation
- Separate the Green and Secondary buttons into shared widgets


lib/features/children/add_member/pages/add_member_form_page.dart
```diff
@@ -6,6 +6,8 @@ import 'package:givt_app/features/children/add_member/widgets/add_member_form.da
 import 'package:givt_app/features/children/add_member/widgets/success_add_member_page.dart';
 import 'package:givt_app/features/children/add_member/widgets/vpc_page.dart';
 import 'package:givt_app/l10n/l10n.dart';
+import 'package:givt_app/shared/widgets/custom_green_elevated_button.dart';
```
pavelpavelko, line 9: First comment is based on user testing: the checkbox looks a bit bigger then in the Figma file (both parts I think, but at least green background).
Also, it's not clear shall the animation be looped (perhaps it makes sense from explanatory perspective(in Figma it looks more clear with looped animation I think)?

lib/features/registration/widgets/registered_check_animation.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final size = MediaQuery.sizeOf(context);
+    return TweenAnimationBuilder(
+      duration: const Duration(milliseconds: 1200),
+      curve: Curves.elasticOut,
+      tween: Tween<double>(begin: 0, end: 1),
+      builder: (context, value, child) {
+        leftPosition = size.width * (0.33 - 0.11 * value);
```
pavelpavelko, line 23: perhaps would be better to put these as constants?

lib/features/registration/widgets/registered_check_animation.dart
```diff
@@ -0,0 +1,46 @@
...
+        topPosition = size.width * (0.32 - 0.22 * value);
+        imageSize = size.width * (0.15 + 0.4 * value);
+
+        return Stack(
+          children: [
+            SvgPicture.asset(
+              'assets/images/registered_check_background.svg',
+              height: MediaQuery.of(context).size.height * 0.4,
```
pavelpavelko, line 31: we already have size above

lib/shared/widgets/custom_green_elevated_button.dart
```diff
@@ -0,0 +1,31 @@
...
+          backgroundColor: AppTheme.givtLightGreen,
+          disabledBackgroundColor: Colors.grey,
+        ),
+        child: Text(
+          title,
+          style: Theme.of(context).textTheme.headlineSmall!.copyWith(
+                color: Colors.white,
+                fontWeight: FontWeight.w900,
+                fontFamily: 'Avenir',
+                fontSize: 18,
```
pavelpavelko, line 25: we do not need to override fontSize. I think our current default approach is to use Theme.of(context).textTheme.{pickYourSizeProperty}, so why do why need to use both approaches at the same time? :)

lib/shared/widgets/custom_secondary_border_button.dart
```diff
@@ -0,0 +1,35 @@
...
+            width: borderWidth.toDouble(),
+          ),
+        ),
+        child: Text(
+          title,
+          style: Theme.of(context).textTheme.headlineSmall!.copyWith(
+                color: AppTheme.givtLightGreen,
+                fontWeight: FontWeight.w900,
+                fontFamily: 'Avenir',
+                fontSize: 18,
```
pavelpavelko, line 29: same here

lib/features/children/add_member/pages/add_member_form_page.dart
Daniela510: Yes you are right, it is bigger than the design. I will make it smaller. And no I don't think its looped. In the ticket it's a gif wich are always looped but it looks strange. But in the figma prototype its just a one time animation.
here:
https://www.figma.com/proto/dBOnDJTvfqLwJsK5yvXukj/Givt4Kids---Ongoing-Design?page-id=7420%3A5136&type=design&node-id=8091-18093&viewport=1393%2C819%2C0.29&t=lTLTb3st5FAnqULW-1&scaling=min-zoom&starting-point-node-id=7432%3A25820&mode=design

But looking at it agai, it needs a delay and to be bouncier.

lib/features/registration/widgets/registered_check_animation.dart
Daniela510: 🤔 leftPosition cannot be const because it uses size; separating 0.33 seems a bit silly to me

MaikelStuivenberg (approved):

PR #537, opened 2024-01-11
Improvement: kids 264 and kids 262 add member when family is already set up

Daniela510:
## Description

Depending if the user already has a family set up (created at least one child or adult) we show different text and actions


lib/app/routes/app_router.dart
```diff
@@ -236,13 +236,17 @@ class AppRouter {
...
-              child: const AddMemeberMainScaffold(),
-            ),
-          ),
+              path: Pages.addMember.path,
+              name: Pages.addMember.name,
+              builder: (context, state) {
+                final familyAlreadyExists = state.extra! as bool;
+                return BlocProvider(
+                  create: (_) => AddMemberCubit(getIt()),
+                  child: AddMemeberMainScaffold(
```
MaikelStuivenberg, line 245: type AddMemeber

MaikelStuivenberg (approved):

PR #540, opened 2024-01-14
Feature: kids-305 add download g4k banner to the family overview page

pavelpavelko:
## Description
- Add download g4k floating action button to the family overview page

### DIFF
--- a/lib/features/children/overview/pages/family_overview_page.dart
+++ b/lib/features/children/overview/pages/family_overview_page.dart
@@ -6,6 +6,7 @@ import 'package:givt_app/app/routes/routes.dart';
 import 'package:givt_app/core/enums/amplitude_events.dart';
 import 'package:givt_app/features/children/overview/cubit/family_overview_cubit.dart';
 import 'package:givt_app/features/children/overview/widgets/children_loading_page.dart';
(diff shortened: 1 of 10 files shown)


Daniela510 (commented):
We already have "Download Givt4Kids" localized, but i guess it's not a problem to have it twice

lib/features/children/overview/pages/family_overview_page.dart
```diff
@@ -89,6 +90,12 @@ class FamilyOverviewPage extends StatelessWidget {
           body: SafeArea(
             child: buildFamilyOverviewBody(state, context),
           ),
+          floatingActionButtonLocation:
+              FloatingActionButtonLocation.centerFloat,
+          floatingActionButton: (state is FamilyOverviewUpdatedState &&
+                  (state.hasChildren || !state.isAdultSingle))
+              ? const DownloadGivtForKidsAppWidget()
+              : null,
```
Daniela510, line 98: The banner is shown even if the user set up children and parents. You can see it in design. It should show if the user clicked "I will do this later" on the success screen or if the user dismissed it manually. Probably needs some local flag i guess 🤔 
Overview

Daniela510 (approved):

Daniela510 (commented):
I see that i read that turnery operator wrong. But the logic to only show the banner when its needed is now added. Nice solution!

PR #543, opened 2024-01-17
Bug: kids-226 Donation history needs to be refreshed when user taps outside of the dialog

pavelpavelko:
## Description
- Close button added to the donation confirmation dialog; 
- Error image added for the error state dialog; 
- The dialog is not dismissible by pressing outside of it anymore
- The dialog size is constant now
- Start refreshing right after decision made (before confirmation page appears) in background
- Add close button to the confirmation pages (approved, declined, error)


Daniela510 (approved):
Works great!

PR #555, opened 2024-01-29
Improvement: Kids-354 Update Family Tiles

Daniela510:
## Description

- Sort adults so that the Givt account owner is first (by name)
- Update the adult row to use tiles
- Update the tile colors
- Adult tiles send a different amplitude event and route to edit profile.
- Remove monster colors from profiles
- Replace monster color on edit-child-profile-page with the tile color.


lib/features/children/overview/widgets/family_available_page.dart
```diff
@@ -14,15 +15,26 @@ class FamilyAvailablePage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final currentUserFirstName = context.read<AuthCubit>().state.user.firstName;
+    // The Givt user profile is first in the list
+    final sortedAdultProfiles = profiles.where((p) => p.isAdult).toList()
```
pavelpavelko, line 20: shall we rather incapsulate such logic (with getter?) somewhere in cubit/state level instead of widget/ui?

lib/features/children/overview/widgets/family_available_page.dart
```diff
@@ -14,15 +15,26 @@ class FamilyAvailablePage extends StatelessWidget {
...
-        ParentOverviewWidget(
-          profiles: profiles.where((p) => p.type == 'Parent').toList(),
+        ProfilesOverviewWidget(
+          profiles: sortedAdultProfiles,
         ),
         const SizedBox(height: 20),
-        ChildrenOverviewWidget(
-          profiles: profiles.where((p) => p.type == 'Child').toList(),
+        ProfilesOverviewWidget(
+          profiles: profiles.where((p) => p.isChild).toList(),
```
pavelpavelko, line 37: same here: what about _children_ getter (with filtering inside) in the cubit?

lib/features/children/overview/widgets/profile_overview_tile.dart
```diff
@@ -0,0 +1,142 @@
...
+          side: BorderSide(
+              width: 2,
+              color: _getBorderColor(
+                context,
+                isGivtAccount,
+              )),
+          backgroundColor: _getBackgroundColor(context),
+        ),
+        onPressed: () => _onTap(context, isGivtAccount),
+        child: SizedBox(
```
pavelpavelko, line 46: why do we need SizedBox here?

lib/features/children/overview/widgets/profile_overview_tile.dart
```diff
@@ -0,0 +1,142 @@
...
+    }
+    if (profile.isAdult) {
+      return AppTheme.givtDisabledBorderBlue;
+    }
+    if (profile.isChild) {
+      return AppTheme.givtLightGreen;
+    }
+    return Theme.of(context).colorScheme.onError;
+  }
+}
```
pavelpavelko, line 142: nice and clear Widget!

lib/features/children/overview/widgets/family_available_page.dart
Daniela510: ah since the widget was getting the list of profile instead of state i overlooked those, oops

lib/features/children/overview/widgets/family_available_page.dart
Daniela510: great idea!

lib/features/children/overview/widgets/profile_overview_tile.dart
Daniela510: No clue actually its removed, and all works the same😅

lib/features/children/overview/widgets/profile_overview_tile.dart
Daniela510: Its just a more generalised version of the children_overview_item, so kudos to you too 😄

pavelpavelko (approved):
lgtm)

PR #570, opened 2024-02-09
Feature: kids-422 Implement VPC error flow 1a

pavelpavelko:


lib/features/children/cached_members/cubit/cached_members_cubit.dart
```diff
@@ -0,0 +1,82 @@
...
+      unawaited(
+        AnalyticsHelper.logEvent(
+          eventName: AmplitudeEvents.memberCreatedSuccesfully,
+          eventProperties: {
+            'nrOfMembers': members.length,
+            'memberNames': memberNames,
+          },
+        ),
+      );
+    } catch (error) {
```
MaikelStuivenberg, line 67: Can we also log the stacktrace? Like in most other places :)
it's something like
```

catch (error, stacktrace) {
 await log.....(e.toString(), stacktrace.toString()
 }
```

MaikelStuivenberg (approved):

PR #578, opened 2024-02-12
Pending allowance item

Daniela510:
## Description

- Check allowances for status and number of attempts


lib/features/children/family_history/models/allowance.dart
```diff
@@ -24,6 +28,22 @@ class Allowance extends HistoryItem {
       type: HistoryTypes.values.firstWhere(
         (element) => element.value == map['donationType'],
       ),
+      status: AllowanceStatus.values.firstWhere(
```
pavelpavelko, line 31: I would suggest to encapsulate such search and put it to the AllowanceStatus enum. Also perhaps add error handling (when filed is empty or has another value)?

lib/features/children/family_history/widgets/allowance_item_widget.dart
```diff
@@ -12,28 +12,37 @@ class AllowanceItemWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     final size = MediaQuery.sizeOf(context);
     final locals = context.l10n;
+    final isNotSuccessful = allowance.status != AllowanceStatus.proccessed &&
```
pavelpavelko, line 15: can we encapsulate the code as a getter in the Allowance class?

pavelpavelko (approved):
looks good but with few small suggestions

PR #579, opened 2024-02-12
Feature/kids 345 allowance not successful dialog

Daniela510:
## Description

- await error even if status is 200
- add new addMemberStatus
- pass the error to the family screen
- create alert dialog
- remove parentGuid from fetchProfiles because it is not used

### DIFF
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -463,7 +463,7 @@ class APIService {
         body: decodedBody,
       );
     }
-    return response.statusCode == 200 && !isError;
+    return response.statusCode == 200;
   }
 
   Future<bool> createChild(Map<String, dynamic> body) async {
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -454,7 +454,16 @@ class APIService {
         body: jsonDecode(response.body) as Map<String, dynamic>,
       );
     }
-    return response.statusCode == 200;
+    final decodedBody = jsonDecode(response.body) as Map<String, dynamic>;
+    final isError = decodedBody['isError'] as bool;
+
+    if (response.statusCode == 200 && isError) {
+      throw GivtServerFailure(
+        statusCode: response.statusCode,
+        body: decodedBody,
+      );
+    }
+    return response.statusCode == 200 && !isError;
   }
 
   Future<bool> createChild(Map<String, dynamic> body) async {
@@ -498,7 +507,7 @@ class APIService {
     return response.statusCode == 200;
   }
 
-  Future<List<dynamic>> fetchProfiles(String parentGuid) async {
+  Future<List<dynamic>> fetchProfiles() async {
     final url = Uri.https(
       apiURL,
       '/givtservice/v1/profiles',
--- a/lib/features/children/add_member/cubit/add_member_cubit.dart
+++ b/lib/features/children/add_member/cubit/add_member_cubit.dart
@@ -143,27 +143,24 @@ class AddMemberCubit extends Cubit<AddMemberState> {
 
     emit(state.copyWith(status: AddMemberStateStatus.loading));
     try {
-      final membersAddedSuccessfully =
-          await _addMemberRepository.addMembers(members);
+      await _addMemberRepository.addMembers(members);
 
-      if (membersAddedSuccessfully) {
-        emit(state.copyWith(status: AddMemberStateStatus.success));
-        unawaited(
-          AnalyticsHelper.logEvent(
-            eventName: AmplitudeEvents.vpcAccepted,
-          ),
-        );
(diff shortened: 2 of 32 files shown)


lib/core/network/api_service.dart
```diff
@@ -454,7 +454,16 @@ class APIService {
         body: jsonDecode(response.body) as Map<String, dynamic>,
       );
     }
-    return response.statusCode == 200;
+    final decodedBody = jsonDecode(response.body) as Map<String, dynamic>;
+    final isError = decodedBody['isError'] as bool;
+
+    if (response.statusCode == 200 && isError) {
```
pavelpavelko, line 460: such combination (code 200 and isError true) looks really strange for me

lib/core/network/api_service.dart
```diff
@@ -454,7 +454,16 @@ class APIService {
...
+      throw GivtServerFailure(
+        statusCode: response.statusCode,
+        body: decodedBody,
+      );
+    }
+    return response.statusCode == 200 && !isError;
```
pavelpavelko, line 466: I think we can get rid from return value since we use Exceptions

lib/features/children/add_member/cubit/add_member_cubit.dart
```diff
@@ -143,36 +143,49 @@ class AddMemberCubit extends Cubit<AddMemberState> {
...
-
-      unawaited(
-        AnalyticsHelper.logEvent(
-          eventName: AmplitudeEvents.vpcAccepted,
-        ),
-      );
+      final membersAddedSuccessfully =
+          await _addMemberRepository.addMembers(members);
+
+      if (membersAddedSuccessfully) {
```
pavelpavelko, line 149: let's get rid from membersAddedSuccessfully - if we are here then there was no exception

lib/features/children/add_member/cubit/add_member_cubit.dart
```diff
@@ -143,36 +143,49 @@ class AddMemberCubit extends Cubit<AddMemberState> {
...
+              'memberNames': memberNames,
+            },
+          ),
+        );
+      }
     } catch (error) {
       await LoggingInfo.instance.error(error.toString());
-      unawaited(
-        AnalyticsHelper.logEvent(
-          eventName: AmplitudeEvents.failedToCreateMember,
```
pavelpavelko, line 169: I think we need to send the event not only for VPC_NOT_SUCCESSFUL

lib/features/children/add_member/repository/add_member_repository.dart
```diff
@@ -2,20 +2,21 @@ import 'package:givt_app/core/network/api_service.dart';
...
   @override
-  Future<void> addMembers(List<Member> members) async {
+  Future<bool> addMembers(List<Member> members) async {
     final profilesJsonList = members.map((member) => member.toJson()).toList();
     final body = {
       'profiles': profilesJsonList,
     };
 
-    await apiService.addMember(body);
+    final isSuccess = await apiService.addMember(body);
```
pavelpavelko, line 19: same here - why do we need return value? Do we use it somewhere?

lib/features/children/overview/widgets/allowance_warning_dialog.dart
```diff
@@ -0,0 +1,81 @@
...
+                ),
+          ),
+          const SizedBox(height: 16),
+          Text(
+            context.l10n.noWorriesWeWillTryAgain,
+            textAlign: TextAlign.center,
+            style: Theme.of(context).textTheme.labelMedium?.copyWith(
+                fontFamily: 'Raleway',
+                fontWeight: FontWeight.w600,
+                color: Color(0xFF617793)),
```
pavelpavelko, line 49: let's move the color to the AppTheme

lib/core/network/api_service.dart
Daniela510: Its what we get from BE, technically the request is successful, accounts are created, but the isError field is true bc allowances are not successful

lib/features/children/add_member/cubit/add_member_cubit.dart
Daniela510: I'll change the amplitude event to "failed_to_get_vpc", and put the more generic failedToCreateMember event to the catch all error.

lib/features/children/add_member/repository/add_member_repository.dart
Daniela510: I used it to make sure the members were added successfully, but as you pointed out that is not necessary. So remoooved

lib/core/network/api_service.dart
pavelpavelko: yeah, I got the point) however it still looks like a hack from tech perspective :)

pavelpavelko (approved):
lgtm)

PR #590, opened 2024-02-15
Feature: kids-466 family goal widget

Daniela510:


lib/features/children/family_goal_tracker/pages/family_goal_tracker.dart
```diff
@@ -0,0 +1,55 @@
...
+          }
+        },
+        builder: (context, state) {
+          if (state.status == GoalTrackerStatus.loading) {
+            return const Padding(
+              padding: EdgeInsets.symmetric(vertical: 20, horizontal: 32),
+              child: Center(child: CircularProgressIndicator()),
+            );
+          }
+          if (state.status == GoalTrackerStatus.noGoalSet) {
```
MaikelStuivenberg, line 41: Maybe a switch would be better here? :)

MaikelStuivenberg (approved):

PR #600, opened 2024-02-21
Feat: adds Makefile support

thassio-vinicius:
## Description
With features such as flavors and Melos, I've noticed that the app contains several commands that can be shortened with a Makefile. This PR provides a simple quality-of-life feature for app contributors and incentivizes more frequent usage of other commands like `make lint`, which may assist with keeping track and following Flutter's linter rules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,30 @@
+.PHONY: lint
+lint:
+	dart format lib test
+	flutter analyze
+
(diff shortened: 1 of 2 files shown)


MaikelStuivenberg (approved):
Thanks! I'm gonna merge this into our repo and also use it for the other app we have :)

PR #622, opened 2024-03-12
Feature/kids 500/kids 620 and kids 507 and kids 504 cubit provider and UI

Daniela510:
## Description

- Add the goal card to the home screen if there is an active goal
- Add goal list item
- Make the goal widgets self-populating

I understood the core issue of providing a cubit to multiple pages is potentially having duplicates in the same context. I could solve this by a higher order cubit, but the only 'elegant' solution is to provide it at the same level as AuthCubit. Could also pass an instance of GoalTracker Cubit through the routes but that also feels hacky. 

In the end I think providing the cubit at the lowest possible level is a good idea. There will be no duplicate cubits in the same context, and the widget will always get the latest info on build.


Daniela510:
Upside of a higher order provider would be that the goal would have a visible 'loading' state every time it has to be displayed. But we could also solve that by caching maybe 🤔 

https://github.com/givtnl/givt-app/assets/43034761/2a8ff7f5-89ad-443d-81db-089c8f8b5c5e

MaikelStuivenberg (approved):

PR #627, opened 2024-03-15
Feature: kids-508 Give to the family goal 

pavelpavelko:
## Description
- switched to the new api to fetch family goal (/latest instead of list)
- family goal widget is hidden from enter amount bottom sheet (but keep in home screen)
- other options of giving to family goal are hidden
- family goal id is passed to the transaction

### DIFF
--- a/lib/core/network/api_service.dart
+++ b/lib/core/network/api_service.dart
@@ -424,7 +424,7 @@ class APIService {
       body: jsonEncode(body),
       headers: {
         'Content-Type': 'application/json',
-        'x-json-casing': 'PascalKeeze'
+        'x-json-casing': 'PascalKeeze',
       },
     );
     if (response.statusCode >= 300) {
@@ -644,7 +644,7 @@ class APIService {
       url,
       headers: {
         'Content-Type': 'application/json',
-        'x-json-casing': 'PascalKeeze'
+        'x-json-casing': 'PascalKeeze',
       },
     );
     if (response.statusCode >= 400) {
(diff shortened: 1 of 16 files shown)


lib/features/children/family_goal_tracker/repository/goal_tracker_repository.dart
```diff
@@ -13,14 +13,14 @@ class GoalTrackerRepositoryImpl with GoalTrackerRepository {
...
   @override
-  Future<List<FamilyGoal>> fetchFamilyGoal() async {
+  Future<FamilyGoal> fetchFamilyGoal() async {
     final response = await _apiService.fetchFamilyGoal();
-
-    final List<FamilyGoal> goals = [];
-
-    for (final item in response) {
-      goals.add(FamilyGoal.fromMap(item as Map<String, dynamic>));
+    FamilyGoal goal;
```
MaikelStuivenberg, line 18: 18 till 24 can be simplified:

`if(response.isEmpty)
return FamilyGoal.empty();

return FamilyGoal.fromMap(response);`

lib/core/network/api_service.dart
```diff
@@ -969,7 +969,7 @@ class APIService {
       );
     }
     final decodedBody = jsonDecode(response.body);
-    final itemMap = decodedBody['items']! as List<dynamic>;
+    final itemMap = decodedBody['item']! as Map<String, dynamic>;
```
MaikelStuivenberg, line 972: It's not really needed to put it in a variable (itemMap) first. return can also be on this line

MaikelStuivenberg (approved):

PR #635, opened 2024-03-19
Refresh family goal after approve/decline donations (KIDS-674)

MaikelStuivenberg:

### DIFF
--- a/lib/features/children/parental_approval/widgets/parental_approval_dialog_content.dart
+++ b/lib/features/children/parental_approval/widgets/parental_approval_dialog_content.dart
@@ -26,6 +26,9 @@ class ParentalApprovalDialogContent extends StatelessWidget {
   Future<void> _refreshHistory(BuildContext context) async {
     await context.read<FamilyOverviewCubit>().fetchFamilyProfiles();
     if (!context.mounted) return;
+    
+    await context.read<GoalTrackerCubit>().getGoal();
+    if (!context.mounted) return;
 
     // Exectue the following cubits in parallel
     await Future.wait([


lib/features/children/parental_approval/widgets/parental_approval_dialog_content.dart
```diff
@@ -26,6 +26,9 @@ class ParentalApprovalDialogContent extends StatelessWidget {
   Future<void> _refreshHistory(BuildContext context) async {
     await context.read<FamilyOverviewCubit>().fetchFamilyProfiles();
     if (!context.mounted) return;
+    
+    await context.read<GoalTrackerCubit>().getGoal();
```
Daniela510, line 30: I would also remove it from the "parallel" below

lib/features/children/parental_approval/widgets/parental_approval_dialog_content.dart
```diff
@@ -26,6 +26,9 @@ class ParentalApprovalDialogContent extends StatelessWidget {
...
+    if (!context.mounted) return;
 
     // Exectue the following cubits in parallel
     await Future.wait([
```
Daniela510, line 34: I would also remove getGoal from the "parallel" below, to not have it twice in this method

MaikelStuivenberg: Now you say this i'm actually wondering why that one doesnt work.. I'll do some more tests!

Daniela510: also why  is `if (!context.mounted) return;` twice also

MaikelStuivenberg: https://dart.dev/tools/linter-rules/use_build_context_synchronously :)

PR #646, opened 2024-03-27
Improve donating from external link (kids-395)

Daniela510:
## Description

 - After completing a donation from external link user will be redirected to the url provided


lib/features/give/pages/giving_page.dart
```diff
@@ -55,12 +56,22 @@ class _GivingPageState extends State<GivingPage> {
...
       Pages.home.name,
       queryParameters: {
         'given': 'true',
       },
     );
+
+    if (afterGivingRedirection.isNotEmpty) {
+      final url = Uri.parse(afterGivingRedirection);
+      if (!await launchUrl(url)) {
+        throw Exception('Could not launch $url');
```
MaikelStuivenberg, line 72: What happens if we throw this exception? 
(I think we should redirect the user back to the normal giving page here)

Daniela510: I'm redirecting to home screen before this, so in any case if the url launches or not the user goes to the home screen. 
In case it's not launched i dont think we need to display anything to the user, but maybe some logging would be good for us to know.

MaikelStuivenberg (approved):

PR #659, opened 2024-04-05
The new biometricCheck AuthStatus and logic based on it is implemented (KIDS-748)

pavelpavelko:

### DIFF
--- a/lib/features/auth/cubit/auth_cubit.dart
+++ b/lib/features/auth/cubit/auth_cubit.dart
@@ -4,6 +4,7 @@ import 'dart:io';
 import 'package:bloc/bloc.dart';
 import 'package:equatable/equatable.dart';
 import 'package:firebase_messaging/firebase_messaging.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_timezone/flutter_timezone.dart';
 import 'package:givt_app/core/enums/enums.dart';
 import 'package:givt_app/core/failures/failures.dart';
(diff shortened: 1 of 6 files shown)


Daniela510 (approved):

lib/features/auth/cubit/auth_cubit.dart
```diff
@@ -47,11 +53,16 @@ class AuthCubit extends Cubit<AuthState> {
...
         state.copyWith(
-          status: AuthStatus.authenticated,
+          status: biometricSetting == BiometricSetting.unknown
+              ? AuthStatus.biometricCheck
+              : AuthStatus.authenticated,
           user: userExt,
           session: session,
+          navigate: navigate,
         ),
       );
```
Daniela510, line 67: Styling opinion but i would wrap the emit in a guarded if statement like 
If (biometricSetting == BiometricSetting.unknown){
emit(
        state.copyWith(
          status: AuthStatus.biometricCheck,
          user: userExt,
          session: session,
          navigate: navigate,
        ),
      );
      
      return;
      }
      
      I think its easier to follow, but its just an opinion 🦐

lib/features/auth/cubit/auth_cubit.dart
```diff
@@ -101,6 +112,16 @@ class AuthCubit extends Cubit<AuthState> {
     }
   }
 
+  void completeBiometricsCheck() =>
+      emit(state.copyWith(status: AuthStatus.authenticated));
+
+  void clearNavigation() => emit(
+        state.copyWith(
+          status: state.status,
+          navigate: AuthState._emptyNavigate,
```
Daniela510, line 121: Curious, why does _emptyNavigate have an underscore, when it isnt private 🤔

lib/features/auth/cubit/auth_cubit.dart
pavelpavelko: perhaps, but in this case we will have two emits, so not sure what is better)

lib/features/auth/cubit/auth_cubit.dart
Daniela510: ohh never digested that cubit and state share private classes, cool!

PR #664, opened 2024-04-08
Accept an invitation to an impact group (for existing users KIDS-724, temp users KIDS-725, and new users KIDS-722)

Daniela510:
## Description

- Impact groups are only fetched for US users
- New & Temp users are redirected to registration
- New & Temp users that are invited do not get the alert dialog, just the invitation bottomsheet
- Existing users are redirected to a success screen that then redirected to the family screen

Important: Api data is mocked


lib/core/enums/amplitude_events.dart
```diff
@@ -2,6 +2,7 @@ enum AmplitudeEvents {
   letsGo('lets_go'),
   login('login'),
   familyClicked('family_clicked'),
+  clickedSeeMyFamily('clicked_see_my_family'),
```
pavelpavelko, line 5: perhaps would be more consistant seeMyFamilyClicked?)

lib/core/network/api_service.dart
```diff
@@ -937,6 +937,65 @@ class APIService {
...
+    // final decodedBody = jsonDecode(response.body) as Map<String, dynamic>;
+    // final itemMap = decodedBody['items']! as List<dynamic>;
+    // return itemMap;
+    final mockData = jsonEncode({
+      "items": [
+        {
+          'id': "GUIDGUIDGUID",
+          "status": "Invited",
+          "name": "Luza Family",
+          "type": "Family",
```
pavelpavelko, line 990: shall we create enum for type?

lib/features/impact_groups/cubit/impact_groups_cubit.dart
```diff
@@ -0,0 +1,79 @@
...
+          await _impactGroupInviteRepository.fetchImpactGroups();
+
+      emit(
+        state.copyWith(
+          status: ImpactGroupCubitStatus.fetched,
+          impactGroups: impactGroups,
+        ),
+      );
+    } catch (e) {
+      emit(
```
pavelpavelko, line 38: loggit?

lib/features/impact_groups/cubit/impact_groups_cubit.dart
```diff
@@ -0,0 +1,79 @@
...
+          error: e.toString(),
+        ),
+      );
+    }
+  }
+
+  void checkForInvites() {
+    for (var i = 0; i < state.impactGroups.length; i++) {
+      if (state.impactGroups[i].status == ImpactGroupStatus.invited) {
+        emit(
```
pavelpavelko, line 50: just wondering how the logic works for multiple invitation at the same time?

lib/features/impact_groups/cubit/impact_groups_cubit.dart
```diff
@@ -0,0 +1,79 @@
...
+    required String groupId,
+  }) async {
+    emit(state.copyWith(status: ImpactGroupCubitStatus.loading));
+    try {
+      await _impactGroupInviteRepository.acceptGroupInvite(
+        groupId: groupId,
+      );
+      await fetchImpactGroups();
+    } catch (e) {
+      emit(
```
pavelpavelko, line 71: loggit

lib/features/impact_groups/impact_group_join_success_page.dart
```diff
@@ -0,0 +1,146 @@
...
+                        );
+                      }),
+                )
+              else
+                Column(
+                  children: [
+                    const DownloadG4KButton(),
+                    TextButton(
+                      onPressed: () {
+                        context.pushReplacementNamed(
```
pavelpavelko, line 123: shall we track amplitude here as well?

lib/features/impact_groups/impact_group_recieve_invite_sheet.dart
```diff
@@ -0,0 +1,86 @@
...
+                        queryParameters: {
+                          'email': user.email,
+                          'createStripe': createStripe.toString(),
+                        },
+                      )
+                      ..pop();
+                  } else {
+                    context
+                      ..goNamed(Pages.joinImpactGroupSuccess.name)
+                      ..pop();
```
pavelpavelko, line 76: I do not understand why do we need this pop() right after goNamed?

Daniela510: It closes the bottomsheet after navigating. If you remove the pop then swipe back you can see the bottomsheet open over the homescreen.

lib/features/impact_groups/cubit/impact_groups_cubit.dart
Daniela510: Only one bottomsheet will pop up, the next one will pop up the next time they go to home page or they logout/login

PR #665, opened 2024-04-09
Generosity challenge closed loop is implemented (KIDS-751, KIDS-750, KIDS-755)

pavelpavelko:
## Description
- Generosity challenge skeleton is implemented; 
- Logic that activates challenge via secret link is implemented; 
- Logic that keeps user in the loop is implemented; 
- Placeholder pages for overview and active day are implemented;

### DIFF
--- a/lib/features/children/generosity_challenge/models/day.dart
+++ b/lib/features/children/generosity_challenge/models/day.dart
@@ -1,0 +1,29 @@
+import 'package:equatable/equatable.dart';
+
+class Day extends Equatable {
+  const Day({
+    required this.dateCompleted,
+  });
+
+  Day.byDateTime({required DateTime dateTime})
+      : dateCompleted = dateTime.toIso8601String();
+
+  const Day.empty() : dateCompleted = '';
+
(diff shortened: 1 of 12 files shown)


lib/features/children/generosity_challenge/models/day.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:equatable/equatable.dart';
+
+class Day extends Equatable {
+  const Day({
+    required this.dateCompleted,
+  });
```
Daniela510, line 6: I imagine that we will expand the Day to also have more challenge data like day title, and daily task/message. What do you think?

lib/features/children/generosity_challenge/models/day.dart
```diff
@@ -0,0 +1,29 @@
...
+
+  final String dateCompleted;
+
+  Map<String, dynamic> toJson() {
+    return {
+      'dateCompleted': dateCompleted,
+    };
+  }
+
+  bool get isCompleted => dateCompleted.isNotEmpty;
```
Daniela510, line 25: nice!

Daniela510 (approved):

PR #670, opened 2024-04-09
Create the generosity challenge overview (kids-753)

Daniela510:
## Description

Simulator Screenshot - iPhone 15 - 2024-04-09 at 16 35 00


lib/features/children/generosity_challenge/pages/generosity_challenge_overview.dart
```diff
@@ -1,39 +1,73 @@
...
+              children: [
+                const SizedBox(height: 24),
+                SvgPicture.asset(
+                  'assets/images/family_avatar.svg',
+                  width: 60,
+                  height: 60,
+                ),
+                const SizedBox(height: 14),
+                Text(
+                  'The ${user.lastName} Family',
```
pavelpavelko, line 41: wondering how {user.lastName} works since generosity challenge should work for unauthenticated users?

lib/features/children/generosity_challenge/widgets/day_item_placeholder.dart
```diff
@@ -1,4 +1,8 @@
 import 'package:flutter/material.dart';
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
+import 'package:givt_app/features/children/generosity_challenge/models/color_combo.dart';
+import 'package:givt_app/shared/widgets/action_container.dart';
+import 'package:givt_app/utils/app_theme.dart';
 
 class DayItemPlaceholder extends StatelessWidget {
```
pavelpavelko, line 7: shall we change the name?))

lib/features/children/generosity_challenge/pages/generosity_challenge_overview.dart
Daniela510: Oh i actually didnt think about that, but i just asked maikel and andy what do we want to do, so if it is an unauthenticated user we wot show that text yet, ill fix it tomorrow

pavelpavelko (approved):

PR #679, opened 2024-04-15
Family Goal giving method is removed (KIDS-781)

pavelpavelko:


Daniela510 (approved):
Reminder: Do not merge until after release

PR #702, opened 2024-04-25
Tab bar navigation issue is fixed (KIDS-895)

pavelpavelko:
## Description
- Ignore PageView's onPageChanged callback when inter-page animation is active


Daniela510 (approved):
Nice!

PR #705, opened 2024-04-25
Switch to groups endpoint as a data source for family goal & overview (kids-889)

Daniela510:
## Description

  Btw goal progress bar was not visually updating after parental approval of donations, making it stateless solved it, thats why the change.


pavelpavelko (approved):
Nice!

PR #711, opened 2024-05-02
Generosity challenge - day 2 - assignment - select family values (kids-923)

Daniela510:
## Description

Overview of changes: 
- Added a DailyAssignmentCubit to keep track of assignments that have a flow between giving assignment and completion. Right now this is specifically about day 2 and day 7.
- From design, day 2 and day 7 when the assignment card is split in 2 parts, the second part is filled in by user choices, on day 2 is the selected values on day 7 it will be a recap of the donation.
- Also looking at the design it seems the banner mage will be the same every day, the top label is also static, and the banner title is the same as the assignment title so i removed the feedback fields as they seemed redundant.
- The family values selection feature is in subfolder assignments & happens on a sub-route of the generosity challenge.


lib/app/routes/pages.dart
```diff
@@ -11,6 +11,10 @@ enum Pages {
     path: '/generosity-challenge',
     name: 'GENEROSITY-CHALLENGE',
   ),
+  generositySelectValues(
+    path: 'select-values',
```
pavelpavelko, line 15: let's add "generosity-..." for consistency

lib/features/children/generosity_challenge/assignments/family_values/models/family_value.dart
```diff
@@ -0,0 +1,36 @@
...
+class FamilyValue {
+  const FamilyValue({
+    required this.displayText,
+    required this.imagePath,
+    required this.colorCombo,
+    required this.interestKeys,
+    required this.area,
+  });
+
+  factory FamilyValue.fromJson(Map<String, dynamic> json) {
```
pavelpavelko, line 12: perhaps fromMap/toMap would be more consistent?

lib/features/children/generosity_challenge/assignments/family_values/pages/select_family_values_page.dart
```diff
@@ -0,0 +1,112 @@
...
+                    backgroundColor: Colors.white,
+                    surfaceTintColor: AppTheme.primary90,
+                    automaticallyImplyLeading: false,
+                    title: Padding(
+                      padding: const EdgeInsets.symmetric(vertical: 24),
+                      child: Row(
+                        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                        children: [
+                          Text(
+                            'Select 3 values',
```
pavelpavelko, line 47: shall we use FamilyValuesState.maxSelectedValues here?

lib/features/children/generosity_challenge/assignments/family_values/repositories/family_values_repository.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:convert';
+
+import 'package:givt_app/features/children/generosity_challenge/assignments/family_values/cubit/family_values_cubit.dart';
+import 'package:givt_app/features/children/generosity_challenge/assignments/family_values/models/family_value.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+
+mixin FamilyValuesRepository {
+  Future<bool> rememberValues({
+    required List<Map<String, dynamic>> body,
```
pavelpavelko, line 9: why do we use map here instead of List<FamilyValue>?

lib/features/children/generosity_challenge/models/task.dart
```diff
@@ -12,21 +12,18 @@ class Task extends Equatable {
     required this.description,
     required this.onTap,
     required this.buttonText,
-    required this.feedbackImage,
```
pavelpavelko, line 15: I would suggest to keep feedback per day logic since it's already implemented and can be useful for upcoming challenges, even if we do not use it for the current one (I assume bc of simplification).

lib/features/children/generosity_challenge/pages/generosity_challenge_day_details.dart
```diff
@@ -20,79 +23,109 @@ class GenerosityChallengeDayDetails extends StatelessWidget {
...
+                    );
+
+                    await AnalyticsHelper.logEvent(
+                      eventName:
+                          AmplitudeEvents.generosityChallengeDayCompleted,
+                      eventProperties: {
+                        'day': challenge.state.detailedDayIndex + 1,
+                      },
+                    );
+                    await challenge.completeActiveDay();
```
pavelpavelko, line 99: at this point I wanna ask why don't we use GenerosityChallengeCubit for it? Isn't DailyAssignmentCubit look redundant?

lib/features/children/generosity_challenge/utils/family_values_content_helper.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:givt_app/features/children/generosity_challenge/models/color_combo.dart';
+
+class FamilyValuesContentHelper {
+  static final List<FamilyValue> _values = [
+    const FamilyValue(
+      displayText: 'Everyone deserves\nhelp from disaster',
+      imagePath:
+          'https://givtstoragedebug.blob.core.windows.net/public/cdn/tag-logos/aid.svg',
+      colorCombo: ColorCombo.highlight,
+      interestKeys: ['AFTERADISASTER'],
```
pavelpavelko, line 11: let's put these values in enums for both interestKeys and area

lib/app/routes/pages.dart
Daniela510: I would reather remove generosity from the title to make it  selectValues(path: 'select-values',) since it is on the generosity sub-route.

lib/features/children/generosity_challenge/assignments/family_values/repositories/family_values_repository.dart
Daniela510: Its of list of maps because its a list of family values, i could wrap the list in a map to be something like:
```
{"values" : [
    { ...
    },
    { ...
    },
  ]
}
```
But i am not sure what would be the utility of that?

lib/features/children/generosity_challenge/pages/generosity_challenge_day_details.dart
Daniela510: I wanted to use smaller, more 'focused', but youu are right, one more GenerosityChallengeStatus would work as well

Daniela510: done

lib/app/routes/pages.dart
pavelpavelko: makes sense bc of sub route, I didn't think about it)

lib/features/children/generosity_challenge/assignments/family_values/repositories/family_values_repository.dart
pavelpavelko: 'Its of list of maps because its a list of family values' - yes, and that's ok. I think repository is a good place for _data translation_ from model to map and to json string and backward instead of doing it in the cubit

pavelpavelko: I mean this part:
`final selectedValuesJson =
        state.selectedValues.map((e) => e.toMap()).toList();`

Daniela510: Ooh! Yes totally, thank you for the direction 🙏

pavelpavelko (approved):
Good job Daniela!

PR #718, opened 2024-05-07
Save user input during the chat for further registration (KIDS-877)

pavelpavelko:


assets/jsons/generosity_challenge/chat_scripts/chat_script_day_0.json
```diff
@@ -11,7 +11,7 @@
       "next": {
         "type": "textMessage",
         "side": "interlocutor",
-        "text": "Great! Hi Peterson Family! I believe you're exactly what Tulsa needs to turn things around",
+        "text": "Great! Hi lastName Family! I believe you're exactly what Tulsa needs to turn things around",
```
TammiLion, line 14: 1) There's currently a risk that one of the keys that you are using will now be unintenionally replace text that does not need/want replacement. For example one of your keys is "email". Therefore a text like "We will send you an email at..." or "You might receive some emails..." will actually get replaced.

2) Having the enum ChatScriptSaveKey actually makes the use of the json a bit inflexible. It means that for every key we need to save we would have to implement it on the front-end. Ideally we could just change the hosted json files and not need to care about any app changes anymore.

Suggestion: Since you store and retrieve your user data as a Map<String, dynamic> I think both issues could be fixed by making use of a "Generic Template String". 

https://stackoverflow.com/questions/57230201/generic-template-string-like-in-python-in-dart/57231174#57231174
https://gist.github.com/creativecreatorormaybenot/7c57666a53cbf21f2adab08b4ad84ef5

In the json, the part you want to replace, for example "lastName" would become "{lastName}". You can just delete the ChatScriptSaveKey class. In _formatChatTextWithUserData you would then do:

return TemplateString(source).format(userData) instead of the for loop you have now.

pavelpavelko: @TammiLion , thank you for the review!
Good point about the key words in texts, thanks. They are with brackets {} now.
Re TemplateString - I’ve found a similar https://pub.dev/packages/format package and the formatting part definitely looks more compact now.
I’m not sure about getting rid of ChatScriptSaveKey enum, since it was the point to have specific list of values we gonna use for both user registration and family profiles creation.
Amplitude tracking for the chat is also fixed (kids-962)

lib/features/children/generosity_challenge/cubit/generosity_challenge_cubit.dart
```diff
@@ -101,6 +103,22 @@ class GenerosityChallengeCubit extends Cubit<GenerosityChallengeState> {
     _refreshState(days: newDays);
   }
 
+  Future<void> saveUserData(ChatScriptItem item) =>
+      _generosityChallengeRepository.saveUserData(
+        ChatScriptSaveKey.fromString(item.saveKey),
```
TammiLion, line 108: Just the item.savekey will suffice as the first parameter if you fix the other comment. Then the ChatScriptSaveKey class can be deleted.

lib/features/children/generosity_challenge_chat/chat_scripts/cubit/chat_scripts_cubit.dart
```diff
@@ -126,19 +153,21 @@ class ChatScriptsCubit extends Cubit<ChatScriptsState> {
...
-            await AnalyticsHelper.logChatScriptEvent(
-              eventName: amplitudeEvent,
-              eventProperties: {
-                'value': state.gainedAnswer.answerText,
-              },
-            );
+          final saveKey =
+              ChatScriptSaveKey.fromString(state.gainedAnswer.saveKey);
+          if (saveKey.isSupported) {
+            await _challengeCubit.saveUserData(state.gainedAnswer);
```
TammiLion, line 166: No checking for supported anymore, just an isEmpty check and then await _challengeCubit.saveUserData(state.gainedAnswer.saveKey);

pubspec.yaml
```diff
@@ -38,6 +38,7 @@ dependencies:
   flutter_svg: ^2.0.9
   flutter_timezone: ^1.0.8
   font_awesome_flutter: ^10.6.0
+  format: ^1.5.1
```
TammiLion, line 41: Nice find! 💪

TammiLion (approved):

assets/jsons/generosity_challenge/chat_scripts/chat_script_day_0.json
TammiLion: Understood!

PR #741, opened 2024-05-10
Extract allowance counter as a reusable widget. 

TammiLion:
Currently this widget is only being used in the add member page. I want to use this same widget on the new edit monthly giving allowance screen.


Daniela510 (approved):
nice

PR #744, opened 2024-05-14
Fix generosity repo crashing on loading from cache.

TammiLion:
Should fix:

https://console.firebase.google.com/u/0/project/givtapp-ebde1/crashlytics/app/android:net.givtapp.droid2/issues/f0e51be3c8ddbb9304c9043cb41d9d6d?time=last-twenty-four-hours&versions=4.2.14%20(483)&sessionEventKey=6641E0BB00EA00033277DE639BB34797_1946863832000665782

### DIFF
--- a/lib/features/children/generosity_challenge/repositories/generosity_challenge_repository.dart
+++ b/lib/features/children/generosity_challenge/repositories/generosity_challenge_repository.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:developer';
 import 'package:flutter/foundation.dart';
 import 'package:givt_app/features/children/generosity_challenge/models/day.dart';
 import 'package:givt_app/features/children/generosity_challenge/utils/generosity_challenge_helper.dart';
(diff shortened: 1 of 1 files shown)


lib/features/children/generosity_challenge/repositories/generosity_challenge_repository.dart
```diff
@@ -35,25 +40,37 @@ class GenerosityChallengeRepositoryImpl with GenerosityChallengeRepository {
...
+              (day) => Day.fromMap(day as Map<String, dynamic>),
+            )
+            .toList();
+        return result;
+      } else {
+        return await _saveAndReturnEmptyDays();
+      }
+    } catch (e, s) {
+      if (kDebugMode) {
+        print(e);
```
pavelpavelko, line 58: shall we use log instead of print here?

pavelpavelko (approved):
lgtm :) 
looks like the reason of the crash was recent extension of the Day model, so that should be fixed

lib/features/children/generosity_challenge/repositories/generosity_challenge_repository.dart
TammiLion: Yes!

PR #745, opened 2024-05-14
Feature/kids 838 implement edit allowance screen

TammiLion:
[## Description

](https://linear.app/givt/issue/KIDS-838/implement-edit-allowance-screen)
https://linear.app/givt/issue/KIDS-840/remove-allowance-input-field-from-edit-profile-screen

So architecturally I kept this PR almost exactly the same as the current Givt app. I am building on the previous PR with the UIModel and having a "dumb" success screen. And in this PR I also illustrate a "dumb" edit allowance page. It's not wired to the cubit, it truly doesn't know anything, it just returns the result of the number that was set by the user when they click confirm (a result is not returned when back is pressed). I also added some convenient extensions so please check them out for future use.


ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -672,11 +672,10 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = "AppIcon-dev";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_IDENTITY = "Apple Development";
```
MaikelStuivenberg, line 675: I assume this is only for local dev and not the build versions of the app? :) As they need to be manual with fastlane

lib/core/network/api_service.dart
```diff
@@ -505,6 +506,27 @@ class APIService {
     return response.statusCode == 200;
   }
 
+  Future<bool> editChildAllowance(String childGUID, int allowance) async {
+    final url =
+        Uri.https(apiURL, '/givtservice/v1/ChildProfile/$childGUID/allowance');
```
MaikelStuivenberg, line 511: Minor backend change. Can you change 'childprofile' to 'profile' in the url? I just made some changes with Pedro to streamline things
Thanks 😇

 /givtservice/v1/profile/$childGUID/allowance

test/golden/shared/pages/goldens/success_pages_golden.png
MaikelStuivenberg, line 1: Doesn't the left example looks better accoording to design that right? :-)

lib/shared/widgets/extensions/route_extensions.dart
```diff
@@ -0,0 +1,41 @@
+import 'dart:io';
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+extension RouteExtensions on Widget {
+  Route<dynamic> toRoute(BuildContext context) {
+    if (Platform.isIOS) {
```
MaikelStuivenberg, line 8: This specific page router is only needed to have a (native) page navigation animaiton, right? :)

We talked some weeks/months ago about having one behaviour for the app. So for example for spinners we wanna use the material (or custom) one and not native :)  Maybe for this navigation thing we also wanna do that? 🤔

ios/Runner.xcodeproj/project.pbxproj
TammiLion: Only for local dev indeed (debug-development)

test/golden/shared/pages/goldens/success_pages_golden.png
TammiLion: Actually the right uses more values from the figma file.

lib/shared/widgets/extensions/route_extensions.dart
TammiLion: Just so I understand, you want to make sure that between gorouter and this extension we always use the same page animations?

TammiLion: For now it has the exact same default page animation as the rest of the app/ gorouter. So I consider it resolved for now.

lib/core/network/api_service.dart
TammiLion: Done

Daniela510 (approved):

PR #746, opened 2024-05-14
 Undo challenge progress per day is implemented (KIDS-980)

pavelpavelko:
## Description
- on long pressed callback is added to the Day button (dev only)
- undo challenge progress per day (both days status and chat history) logic is implemented
- simple confirmation dialog is implemented
- note: saved user data stays in the prefs and should be overwritten later on by the chat


Daniela510 (approved):
nice!

PR #749, opened 2024-05-16
Bug fix: next available day is after midnight not after 24 hours

TammiLion:
Next available day should become available after midnight not after 24 hours.


lib/features/children/generosity_challenge/cubit/generosity_challenge_cubit.dart
```diff
@@ -195,18 +196,21 @@ class GenerosityChallengeCubit extends Cubit<GenerosityChallengeState> {
...
+          // for development/testing purposes, no need to do an actual time-check
           return nextActiveDayIndex;
         } else {
-          //no active day yet
-          return -1;
+          final lastCompletedDateTime = DateTime.parse(
+            days[lastCompletedDayIndex].dateCompleted,
+          );
+          final nowDateTime = DateTime.now();
+          if (lastCompletedDateTime.day != nowDateTime.day) {
```
MaikelStuivenberg, line 208: Maybe we need to include something like week number in here as well? :)

MaikelStuivenberg (approved):

PR #760, opened 2024-05-22
Feature: Remember allowance  for kids (kids-956)

Daniela510:


lib/features/children/overview/pages/edit_allowance_page.dart
```diff
@@ -56,15 +64,16 @@ class _EditAllowancePageState extends State<EditAllowancePage> {
                       const SizedBox(height: 8),
                       Text(
                         'Which amount should be added to\n'
-                            "your child's wallet each month?",
+                        "your child's wallet each month?",
```
TammiLion, line 67: I think the only thing missing here is that for the generosity challenge it says "Your child(ren)...", maybe we can add a bool isMultiple which exchanges this text? (or any other solution is fine)

lib/features/children/overview/pages/models/edit_allowance_success_uimodel.dart
```diff
@@ -1,5 +1,7 @@
 class EditAllowanceSuccessUIModel {
-  const EditAllowanceSuccessUIModel({this.amountWithCurrencySymbol});
+  const EditAllowanceSuccessUIModel(
+      {this.amountWithCurrencySymbol, this.onClickButton});
```
TammiLion, line 3: I think here as well we have a small text change: "Your child(ren) will receive $15 each month. You can edit this amount any time."

TammiLion (approved):

PR #763, opened 2024-05-23
Feature: top up screen (kids-843)

Daniela510:


lib/features/children/details/cubit/child_details_cubit.dart
```diff
@@ -77,13 +78,67 @@ class ChildDetailsCubit extends Cubit<ChildDetailsState> {
     }
   }
 
+  Future<void> topUp(int amount) async {
+    await _logTopUpEvent(amount);
```
TammiLion, line 82: Small thing. Personally i would not "await" analytics event. So I would always wrap them in unawaited(); My reasoning is we wouldn't want to wait on analytics in the user experience and if analytics fail its not blocking for the app.

lib/features/children/details/cubit/child_details_state.dart
```diff
@@ -15,6 +15,10 @@ final class ChildDetailsFetchingState extends ChildDetailsState {
   const ChildDetailsFetchingState();
 }
 
+final class ChildTupUpFundsErrorState extends ChildDetailsState {
```
TammiLion, line 18: tupup => topup

TammiLion (approved):
Very good, you apply learnings very fast and you split up future tickets nicely 👍

PR #766, opened 2024-05-23
Generosity challenge completion with navigation to the Givt app (KIDS-958, KIDS-993)

pavelpavelko:
## Description
- completeChallenge method added to the GenerosityChallengeCubit; 
- clear user data logic is implemented; 
- prevent user to start challenge if it was already completed logic is implemented; 
- challenge completion trigger with navigation to the Givt app is implemented (kids-993); 
- complete challenge button (for dev builds only) with confirmation dialog is implemented;

### DIFF
--- a/lib/features/children/generosity_challenge/pages/generosity_challenge.dart
+++ b/lib/features/children/generosity_challenge/pages/generosity_challenge.dart
@@ -50,12 +50,21 @@ class _GenerosityChallengeState extends State<GenerosityChallenge>
     super.dispose();
   }
 
+  Future<void> _onCompleteChallenge() async {
+    await context.read<GenerosityChallengeCubit>().completeChallenge();
+    if (mounted) {
+      context.goNamed(Pages.childrenOverview.name);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


lib/features/children/generosity_challenge/pages/generosity_challenge.dart
```diff
@@ -50,12 +50,21 @@ class _GenerosityChallengeState extends State<GenerosityChallenge>
     super.dispose();
   }
 
+  Future<void> _onCompleteChallenge() async {
+    await context.read<GenerosityChallengeCubit>().completeChallenge();
+    if (mounted) {
+      context.goNamed(Pages.childrenOverview.name);
```
TammiLion, line 56: Can you get the navigation stack in order? I think this would just push the children overview page on top of the generosity challenge. But what we actually need is the current navigation stack cleared and then home and then on top of home the childrenoverview.

Bonus: if you can get the "download the kids4gift app screen" to go in between.

TammiLion (approved):

PR #767, opened 2024-05-23
Feature/kids 844 success screen

Daniela510:


lib/features/children/details/pages/child_details_page.dart
```diff
@@ -67,7 +69,17 @@ class ChildDetailsPage extends StatelessWidget {
             builder: (_) => const TopUpFailureDialog(),
           );
         } else if (state is ChildTopUpSuccessState) {
-          // TODO Kids-844
+          Navigator.push(
```
TammiLion, line 72: Nice 👌

lib/features/children/overview/pages/models/top_up_success_uimodel.dart
```diff
@@ -0,0 +1,9 @@
+class TopUpSuccessUIModel {
```
TammiLion, line 1: 🎉

TammiLion: Good work in implementing the UIModel. The only thing I would change is I would remove the "onClickButton" here as it isn't really "UI" but rather "logic". I would pass it as a parameter in the TopUpSuccessPage itself. That way when we use UIModels in our cubits the cubits don't need to know/ define the onClick.

TammiLion (approved):
Looks good, one small comment!

PR #784, opened 2024-06-06
Bugfix: Don't schedule notifications on Android 7 so the app doesnt crash

MaikelStuivenberg:

### DIFF
--- a/lib/core/notification/notification_service.dart
+++ b/lib/core/notification/notification_service.dart
@@ -1,6 +1,8 @@
 import 'dart:convert';
+import 'dart:io';
 import 'dart:math';
 
+import 'package:device_info_plus/device_info_plus.dart';
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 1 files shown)


lib/core/notification/notification_service.dart
```diff
@@ -286,6 +288,15 @@ class NotificationService implements INotificationService {
     id ??= Random().nextInt(9999 - 1000) + 1000;
 
     await setupFlutterNotifications();
+
+    if (Platform.isAndroid) {
+      final androidInfo = await DeviceInfoPlugin().androidInfo;
```
TammiLion, line 293: Do you think it would be good to surround this section with a try/catch in case getting the info fails?

TammiLion (approved):

lib/core/notification/notification_service.dart
MaikelStuivenberg: Can you recheck? :)

MaikelStuivenberg: I think in this way we can catch errors while fetching android info and errors while scheduling the notification. 

In all error cases i'm not scheduling the notification, just to be sure.. :)

PR #798, opened 2024-06-11
Prevent double loading at the same time, make api calls parallel, refetch everything when loading the page

MaikelStuivenberg:
## Description

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit 34d0c1dfb77039983016171d5b230e786586fd39  | 
|--------|--------|

### Summary:
Enhanced the loading and fetching mechanisms for history data in the application, including preventing double loading, making API calls in parallel, and refetching data on page load.

**Key points**:
- Prevent double loading in `HistoryCubit` by checking load status before proceeding.
- Implement parallel API calls in `HistoryCubit` for fetching donation and allowance history.
- Refetch history data from the beginning when loading the `HomeScreen`.
- Minor adjustments in `routes.dart` and `home_screen.dart` to integrate changes.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)


MaikelStuivenberg:
@ellipsis-dev review this

lib/features/family/features/history/history_cubit/history_cubit.dart
```diff
@@ -12,22 +12,35 @@ class HistoryCubit extends Cubit<HistoryState> {
...
+      final allowanceHistoryFuture = historyRepo.fetchHistory(
         childId: childId,
         pageNumber: state.pageNr,
         type: HistoryTypes.allowance,
       );
 
+      // In this way the second future will already start fetching
+      // while waiting for the first one to finish
+      // This will make the fetching faster
+      final donationHistory = await donationHistoryFuture;
```
TammiLion, line 41: I don't think it does what you say it does...

Scherm­afbeelding 2024-06-12 om 09 51 52

If you want to fetch the calls in parallel I would recommend the Future.wait([]) function:
https://api.dart.dev/stable/3.4.3/dart-async/Future/wait.html

so:

`var responses = await Future.wait([donationHistoryFuture, allowanceHistoryFuture])`

TammiLion (approved):
Please take a look at the comment, for the rest it looks good.

lib/features/family/features/history/history_cubit/history_cubit.dart
MaikelStuivenberg: You can have futures without awaiting them :) they also will get executed immediately.

MaikelStuivenberg: so the execution process will start already before the first await

MaikelStuivenberg: I think the future.wait is only needed if you need to be sure at that point that it has both results. But in our case its fine if it doesnt have the second one yet

TammiLion: Good point, indeed nothing needs to be changed then!

PR #838, opened 2024-06-25
Improvements on back button (KIDS-1118)

MaikelStuivenberg:


lib/shared/widgets/buttons/back_button_app_bar.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:flutter/material.dart';
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
+import 'package:go_router/go_router.dart';
+
+class BackButtonAppBar extends StatelessWidget {
```
TammiLion, line 5: This is not an appbar itself right? So I would change the name to BackButtonLeading / BackButtonAppBarLeading

TammiLion (approved):

PR #843, opened 2024-06-26
Feature Allow cancelling the RGA (Kids-1089)

Daniela510:

### DIFF
--- a/lib/features/children/details/cubit/child_details_cubit.dart
+++ b/lib/features/children/details/cubit/child_details_cubit.dart
@@ -124,6 +124,7 @@ class ChildDetailsCubit extends Cubit<ChildDetailsState> {
         //has to be awaited in order for the ui to update properly
         await fetchChildDetails();
       } else {
+        _emitData();
         emit(
           const ChildDetailsErrorState(
             errorMessage: 'Failed to fetch details',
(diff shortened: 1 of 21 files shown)


lib/features/children/details/cubit/child_details_cubit.dart
```diff
@@ -105,6 +105,49 @@ class ChildDetailsCubit extends Cubit<ChildDetailsState> {
...
+    try {
+      _emitLoading();
+      final isSuccess = await _editChildRepository.cancelAllowance(_profile.id);
+      if (isSuccess) {
+        _emitData();
+        emit(const ChildCancelAllowanceSuccessState());
+        //has to be awaited in order for the ui to update properly
+        await fetchChildDetails();
+      } else {
+        emit(
```
TammiLion, line 127: I think that if isSuccess is now false  you'll show the dialog but the screen will stay stuck in the loading state, i think you need the _emitData there as well.

lib/features/children/overview/pages/edit_allowance_page.dart
```diff
@@ -105,11 +115,43 @@ class _EditAllowancePageState extends State<EditAllowancePage> {
...
+                    mainAxisSize: MainAxisSize.min,
+                    children: [
+                      CustomGreenElevatedButton(
+                        title: context.l10n.confirm,
+                        onPressed: () {
+                          Navigator.of(context).pop(_allowance);
+                        },
+                      ),
+                      Visibility(
+                        visible: widget.initialAllowance != null &&
```
TammiLion, line 128: Nice!

TammiLion (approved):
Just one comment to check!

PR #847, opened 2024-06-26
Create kids with a single topup instead of rga (kids-857 & kids-858)

Daniela510:
## Summary by CodeRabbit

- **New Features**
  - Added `allowanceType` to the member adding process to specify the type of allowance.

- **Enhancements**
  - Updated `addMembers` method to include `isRGA` parameter for additional functionality.
  - Modified the form to display a message about starting a child's giving journey.
  - Changed initial allowance value for `AllowanceCounter` to `5`.

- **Bug Fixes**
  - Fixed constructor and updated text in `GivingAllowanceInfoButton`.

### DIFF
--- a/lib/features/children/add_member/repository/add_member_repository.dart
+++ b/lib/features/children/add_member/repository/add_member_repository.dart
@@ -22,6 +22,7 @@ class AddMemberRepositoryImpl with AddMemberRepository {
     final profilesJsonList = members.map((member) => member.toJson()).toList();
     final body = {
       'profiles': profilesJsonList,
+      'allowanceType': 1,
     };
 
     await apiService.addMember(body);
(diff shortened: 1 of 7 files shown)


TammiLion (approved):

lib/features/children/add_member/repository/add_member_repository.dart
```diff
@@ -22,6 +22,7 @@ class AddMemberRepositoryImpl with AddMemberRepository {
     final profilesJsonList = members.map((member) => member.toJson()).toList();
     final body = {
       'profiles': profilesJsonList,
+      'allowanceType': 1,
```
TammiLion, line 25: Can you check if this is also used by the generosity challenge code? If not then approved.

PR #865, opened 2024-07-03
Fix timezone for notifications (kids-1114 and kids-1144)

Daniela510:
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated time zone location from 'America/Tulsa' to 'America/Chicago' for improved regional accuracy.

- **New Features**
  - Introduced asynchronous operations in schedule generation and notification setup for smoother performance.
  - Added support for `FlutterTimezone` to enhance timezone handling.

### DIFF
--- a/lib/features/children/generosity_challenge/utils/generosity_challenge_helper.dart
+++ b/lib/features/children/generosity_challenge/utils/generosity_challenge_helper.dart
@@ -33,7 +33,7 @@ class GenerosityChallengeHelper {
     bool isDebug,
   ) {
     final now = tz.TZDateTime.now(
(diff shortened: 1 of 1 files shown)


lib/features/children/generosity_challenge/utils/generosity_challenge_helper.dart
```diff
@@ -33,7 +33,7 @@ class GenerosityChallengeHelper {
     bool isDebug,
   ) {
     final now = tz.TZDateTime.now(
-      tz.getLocation(isDebug ? 'Europe/Amsterdam' : 'America/Tulsa'),
```
MaikelStuivenberg, line 36: Just curious, why wouldn't we do local time on users device? 

Receiving the notification will then always happen at 4pm no matter where to the user is

PR #866, opened 2024-07-03
Feature: Redirect new users to generosity challenge (kids-1141)

Daniela510:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new screen for redirecting users to the Generosity Challenge registration.
  - Added icons for pending status, warnings, and family superheroes.

- **Enhancements**
  - Improved app navigation logic and redirection handling.
  - Updated `NoChildrenPage` to use a common icon function.

- **Routes**
  - Added a new route for the Generosity Challenge registration redirect.
  - Updated `FamilyPages` enum with a new redirect entry.

- **Events**
  - Added new user interaction events for registration actions.

### DIFF
--- a/lib/app/routes/app_router.dart
+++ b/lib/app/routes/app_router.dart
@@ -647,13 +647,14 @@ class AppRouter {
               GenerosityChallengeHelper.isCompleted)) {
         if (state.user.needRegistration) {
           final createStripe = state.user.personalInfoRegistered;
-          context.pushNamed(
-            FamilyPages.registrationUS.name,
-            queryParameters: {
+          context
+            ..pushNamed(FamilyPages.registrationUS.name, queryParameters: {
               'email': state.user.email,
               'createStripe': createStripe.toString(),
-            },
-          );
+            })
+            ..pushNamed(
+              FamilyPages.generosityChallengeRedirect.name,
+            );
         } else if (routerState.name == Pages.loading.name) {
           context.goNamed(FamilyPages.profileSelection.name);
         }
--- a/lib/app/routes/app_router.dart
+++ b/lib/app/routes/app_router.dart
@@ -646,15 +646,9 @@ class AppRouter {
           (!GenerosityChallengeHelper.isActivated ||
               GenerosityChallengeHelper.isCompleted)) {
         if (state.user.needRegistration) {
-          final createStripe = state.user.personalInfoRegistered;
-          context
-            ..pushNamed(FamilyPages.registrationUS.name, queryParameters: {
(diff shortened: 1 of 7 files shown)


lib/app/routes/app_router.dart
```diff
@@ -647,13 +647,14 @@ class AppRouter {
...
-          context.pushNamed(
-            FamilyPages.registrationUS.name,
-            queryParameters: {
+          context
+            ..pushNamed(FamilyPages.registrationUS.name, queryParameters: {
               'email': state.user.email,
               'createStripe': createStripe.toString(),
-            },
-          );
+            })
```
MaikelStuivenberg, line 654: Why push both and not just one? :)
From the other page you can open the registration page

lib/features/family/app/family_pages.dart
```diff
@@ -64,6 +64,10 @@ enum FamilyPages {
       path: 'choose-amount-slider-goal', name: 'CHOOSE_AMOUNT_SLIDER_GOAL'),
 
   //generosity challenge
+  generosityChallengeRedirect(
+    path: '/generosity-challenge-redirect',
```
MaikelStuivenberg, line 68: Does this need to be a root path or should it be without the slash?

lib/app/routes/app_router.dart
Daniela510: The final design looks like a screen, so why would I open a screen from a screen?
I can open a bottomsheet though from the registration page, is that what you meant?

MaikelStuivenberg: I mean that you are doing 2x 
            ..pushNamed()
Right now.

But I don't understand why :)
You can only do a push to the new one, right?

lib/features/family/app/family_pages.dart
Daniela510: Oh right, ill put it at the same level as registrationUs

lib/app/routes/app_router.dart
Daniela510: I am setting up the stack so that the GivtBackButton in the Appbar would go to the page under. I _can_ only push one, then overwrite the GivtBackButton to navigate instead of pop, but I am not sure what are the benefits of one or the other.

Daniela510: I mean there already is onPressedForced so no hassle, but I would like to know why? 😅

Daniela510: Anw its changed to only push 1 page 🤷‍♀️

MaikelStuivenberg (approved):

PR #868, opened 2024-07-04
Fix: Set user id in Amplitude when new/temp user signs up (kids-444)

Daniela510:
## Summary by CodeRabbit

- **New Features**
  - Introduced analytics tracking for temporary user actions.
  - Added user property setting for new user sign-ups.

- **Enhancements**
  - Improved sign-in flow with additional parameter support for better tracking and analytics.

### DIFF
--- a/lib/features/auth/cubit/auth_cubit.dart
+++ b/lib/features/auth/cubit/auth_cubit.dart
@@ -54,6 +54,17 @@ class AuthCubit extends Cubit<AuthState> {
       );
 
       var userExt = await _authRepositoy.fetchUserExtension(session.userGUID);
+      if (password == TempUser.defaultPassword) {
+        await AnalyticsHelper.setUserProperties(userId: userExt.guid);
+        unawaited(
+          AnalyticsHelper.logEvent(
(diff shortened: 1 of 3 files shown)


lib/features/auth/cubit/auth_cubit.dart
```diff
@@ -54,6 +54,17 @@ class AuthCubit extends Cubit<AuthState> {
       );
 
       var userExt = await _authRepositoy.fetchUserExtension(session.userGUID);
+      if (password == TempUser.defaultPassword) {
+        await AnalyticsHelper.setUserProperties(userId: userExt.guid);
+        unawaited(
+          AnalyticsHelper.logEvent(
+            eventName: AmplitudeEvents.continueByEmailSignUpTempUserClicked,
+            eventProperties: {
+              'country': userExt.countryCode,
```
TammiLion, line 63: Can you make this profile_country instead of country? Amplitude will overwrite the country attribute.

TammiLion (approved):

lib/features/auth/cubit/auth_cubit.dart
Daniela510: Oki 👌

PR #887, opened 2024-07-15
Bug: timer screen back button does not work (KIDS-1197)

TammiLion:
## Description

https://linear.app/givt/issue/KIDS-1197/timer-screen-back-button-on-this-screen-doesnt-work

Also add a test button that sets the remaining duration to 20 seconds for easier testing. Put player in low latency mode for the most of it, syncs the tick tock sounds and vibration better.

## Summary by CodeRabbit

- **New Features**
  - Added a debug button to set the timer to 20 seconds remaining in debug mode.

- **Improvements**
  - Enhanced alarm sound handling for better player setup and completion.
  
- **Developer Tools**
  - Introduced a mechanism to check if the app is in debug mode.

### DIFF
--- a/lib/features/children/generosity_challenge/assignments/day4_timer/pages/day4_timer_screen.dart
+++ b/lib/features/children/generosity_challenge/assignments/day4_timer/pages/day4_timer_screen.dart
@@ -11,9 +11,11 @@ import 'package:givt_app/features/children/generosity_challenge/widgets/generosi
 import 'package:givt_app/features/children/generosity_challenge/widgets/generosity_back_button.dart';
 import 'package:givt_app/features/family/app/family_pages.dart';
 import 'package:givt_app/features/family/helpers/vibrator.dart';
+import 'package:givt_app/shared/widgets/buttons/custom_green_elevated_button.dart';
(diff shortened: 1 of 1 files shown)


lib/features/children/generosity_challenge/assignments/day4_timer/pages/day4_timer_screen.dart
```diff
@@ -112,6 +115,11 @@ class _Day4TimerScreenState extends State<Day4TimerScreen> {
     );
   }
 
+  Future<bool> _isDebugApp() async {
```
MaikelStuivenberg, line 118: Should we put this into a more generic place? I think I'm seen this method more and more in our code :)

MaikelStuivenberg (approved):

lib/features/children/generosity_challenge/assignments/day4_timer/pages/day4_timer_screen.dart
TammiLion: Yea, agreed I'll make a separate PR for it!

PR #892, opened 2024-07-15
Bug/kids 1132 improve family app tracking

TammiLion:
## Description

https://linear.app/givt/issue/KIDS-1132/is-family-app-tracking-seems-off

Now logs user properties on login, profile switch and logout correctly.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to clear user properties on logout.
  - Added a method to extract and log user properties for better event tracking.

- **Enhancements**
  - Improved user property logging across various features for more consistent analytics.
  - Refactored user property handling in the profile selection and credit card details screens for streamlined user tracking.

- **Bug Fixes**
  - Removed redundant analytics calls to enhance performance.


lib/features/auth/cubit/auth_cubit.dart
```diff
@@ -200,14 +200,13 @@ class AuthCubit extends Cubit<AuthState> {
...
 
-    AnalyticsHelper.setFamilyAppTracking(isOn: false);
-
     emit(
       state.copyWith(
         status: AuthStatus.unauthenticated,
         email: state.user.email,
       ),
     );
+    unawaited(AnalyticsHelper.clearUserProperties());
```
MaikelStuivenberg, line 209: If we make a new method. Isn't it better to do the unawaited part in there and not make the method async?
I guess there is never a reason to wait, right?

lib/utils/analytics_helper.dart
```diff
@@ -58,16 +53,30 @@ class AnalyticsHelper {
       eventProperties: eventProperties,
     );
 
-    log('$eventName pressed with properties: $eventProperties');
+    log('$eventName pressed with event properties: $eventProperties');
+  }
+
+  static Future<void> clearUserProperties() async {
```
MaikelStuivenberg, line 59: see previous comment

MaikelStuivenberg (approved):

lib/features/auth/cubit/auth_cubit.dart
TammiLion: Good catch, fully agreed!

PR #908, opened 2024-07-18
Fix: Minor ui fixes: dynamic sizing of icons and text fix

Daniela510:
## Summary by CodeRabbit

- **New Features**
  - Enhanced responsiveness and visual presentation of various widgets and dialogs, adapting sizes dynamically based on the device's screen width.

- **Bug Fixes**
  - Adjusted text in the `CancelAllowanceDialog` to accurately reflect the action for canceling a "Recurring Amount."

- **Improvements**
  - Updated layout and sizing adjustments for multiple components, including `ChildDetailsPage` and related dialogs, ensuring a more consistent user experience across the application.

### DIFF
--- a/lib/features/children/details/widgets/child_cancel_rga_failed_dialog.dart
+++ b/lib/features/children/details/widgets/child_cancel_rga_failed_dialog.dart
@@ -15,7 +15,10 @@ class ChildCancelRGAFailedDialog extends StatelessWidget {
         child: Column(
           mainAxisSize: MainAxisSize.min,
           children: [
-            warningIcon(),
+            warningIcon(
+              width: MediaQuery.of(context).size.width * 0.35,
+              height: MediaQuery.of(context).size.width * 0.35,
+            ),
             const SizedBox(height: 16),
             Text(
               'Oops something went wrong',
--- a/lib/shared/widgets/common_icons.dart
+++ b/lib/shared/widgets/common_icons.dart
@@ -104,7 +104,7 @@ Widget day4TimerIconRed({double? width, double? height}) => SvgPicture.asset(
     );
 Widget primaryCircleWithText({
   String? text = '',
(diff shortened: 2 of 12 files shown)


lib/features/children/details/widgets/child_cancel_rga_failed_dialog.dart
```diff
@@ -15,7 +15,10 @@ class ChildCancelRGAFailedDialog extends StatelessWidget {
         child: Column(
           mainAxisSize: MainAxisSize.min,
           children: [
-            warningIcon(),
+            warningIcon(
```
MaikelStuivenberg, line 18: Are you sure the icon should scale? In figma it looks fixed size :)

lib/shared/widgets/common_icons.dart
```diff
@@ -104,48 +104,56 @@ Widget day4TimerIconRed({double? width, double? height}) => SvgPicture.asset(
...
-      child: Center(
-        child: Icon(
-          iconData,
-          color: iconColor ?? FamilyAppTheme.primary20,
-          size: iconSize,
+    Padding(
+      // 14 is what the design has when exporting svgs
+      padding: const EdgeInsets.all(14),
+      child: Container(
+        width: circleSize - 28,
```
TammiLion, line 145: I would not do that here. I would change the defaultParam to circleSize = 68. Because now if you pass a size you will go cuckoo, like "WHY IS IT NOT THE SIZE I PASSED".

TammiLion (approved):
1 comment

lib/features/children/details/widgets/child_cancel_rga_failed_dialog.dart
Daniela510: We had a call with And y and decided to try this because with fixed size as it was it looked small on some phones

PR #909, opened 2024-07-18
Feature: fix adminfee architecture and display clean architecture example

TammiLion:
## Description

Also fix for: https://linear.app/givt/issue/KIDS-1219/add-kid-flow-when-you-select-0-it-still-shows-admin-fee-030

## Summary by CodeRabbit

- **New Features**
  - Introduced `AdminFeeCubit` for managing administrative fees.
  - Added `AdminFeeRepository` for handling admin fee data retrieval and caching.
  - Implemented `AdminFeeUIModel` for representing admin fee details in the UI.
  - Added `AdminFeeTextLayout` widget for displaying admin fee information with customizable styles and themes.

- **Enhancements**
  - Refactored `AdminFeeText` widget to improve interaction with the newly introduced `AdminFeeCubit`.
  - Updated `BaseStateConsumer` to allow more flexible state management in the UI.


lib/features/family/features/admin_fee/presentation/widgets/admin_fee_text_layout.dart
```diff
@@ -0,0 +1,33 @@
...
+  final bool isMonthly;
+  final bool isMultipleChildren;
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = themeData ?? Theme.of(context);
+    final perchild = isMultipleChildren ? ' per child' : '';
+    final feeString = uiModel.fee.toStringAsFixed(2);
+    final monthlyText = 'Admin fee of \$$feeString applies$perchild monthly';
+    final nonMonthlyText = 'An admin fee of \$$feeString will be added';
```
TammiLion, line 26: Something we need to decide as a team. I said many times I want dumb widgets and to clarify, dumb when it comes to "business" logic. But I think the widget is allowed to be smart about "presentation" logic, i.e. what and how things should look like. Do you guys agree/ disagree?

lib/features/family/features/admin_fee/presentation/widgets/admin_fee_text.dart
```diff
@@ -18,25 +20,21 @@ class AdminFeeText extends StatefulWidget {
...
-    final fee =
-        _adminFeeRepository.getTotalFee(widget.amount).toStringAsFixed(2);
-    final monthlyText = 'Admin fee of \$$fee applies$perchild monthly';
-    final nonMonthlyText = 'An admin fee of \$$fee will be added';
-    return Text(
-      widget.isMonthly ? monthlyText : nonMonthlyText,
-      style: widget.textStyle ??
-          theme.textTheme.bodySmall!.copyWith(fontWeight: FontWeight.w700),
+    return BaseStateConsumer(
+      cubit: getIt<AdminFeeCubit>()..setAmount(amount),
```
TammiLion, line 26: This is the only thing that's hacky/ i'm unhappy about. Normally I would define this as a field and pass it here. Then i would call init/ setAmount from initState/didChangeDependencies. The problem here is that I want this to be a "add and forget about it widget". When the widget this widget is nested in passes a different amount from the initial you can only know this new value via the build method.

2 solutions:
1) they way i've done it now
2) when the top widget changes the amount it should notify the AdminFeeCubit (that's the clean way to do it)

But it means that everywhere the AdminFee widget is being used you also need to notify the cubit and i just want this to be easy to use. Lemme know your thoughts!

lib/features/family/features/admin_fee/domain/models/admin_fee.dart
```diff
@@ -21,6 +23,17 @@ class AdminFee {
     );
   }
 
+  double getTotalFee(double amount) {
```
TammiLion, line 26: My domain model is kinda smart now. And alternate solution would be to create an extension on AdminFee and add these methods there. Opinions?

Daniela510: Well, ultimately the nr we need to display in I is the getTotal fee so why is this not also a getter in the repo as single source of data?

TammiLion: Because to me a Repository responsibility is retrieving and caching the data and being the single source of truth for the "data". But how the fee is calculated is not data but business logic ("rules"). To me it falls under the domain layer. Does that make sense?

Daniela510: Yes it makes sense, then leave it here and lets have calculations like these always in the domain layer

Daniela510 (approved):

PR #915, opened 2024-07-23
Feature: Accept family invite as new user (kids-1223)

Daniela510:
## Description

Changes to the navigation logic in the new user flow

## Summary by CodeRabbit

- **New Features**
  - Enhanced routing based on user authentication status for improved responsiveness.
  - Simplified navigation to profile selection, improving user experience.
  - Added a new property to indicate the presence of family profiles in fetched data.

- **Bug Fixes**
  - Improved navigation logic to dynamically respond based on user family status.

- **Style**
  - Updated text styles and improved code readability across various components.

- **Chores**
  - Refined button functionality to ensure appropriate navigation or logout actions based on the context state.

### DIFF
--- a/lib/app/routes/app_router.dart
+++ b/lib/app/routes/app_router.dart
@@ -686,12 +686,12 @@ class AppRouter {
     GoRouterState routerState,
   ) async {
     if (state.status == AuthStatus.biometricCheck) {
-      context.pushNamed(
+      unawaited(context.pushNamed(
         state.user.isUsUser
             ? FamilyPages.permitUSBiometric.name
             : Pages.permitBiometric.name,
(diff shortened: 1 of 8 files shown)


lib/app/routes/app_router.dart
```diff
@@ -686,12 +686,12 @@ class AppRouter {
     GoRouterState routerState,
   ) async {
     if (state.status == AuthStatus.biometricCheck) {
-      context.pushNamed(
+      unawaited(context.pushNamed(
```
MaikelStuivenberg, line 689: I think this needs to be awaited because we want to wait for the push. 
But if we do it the same in other places you can keep it.

MaikelStuivenberg (approved):

PR #934, opened 2024-08-01
Add generic bottomsheet and implementation for empty wallet situations (KIDS-1238)

MaikelStuivenberg:
image

## Summary by CodeRabbit

- **New Features**
	- Introduced an "Empty Wallet" bottom sheet to guide users when their wallet balance is zero.
	- Added a new customizable bottom sheet widget, `GivtBottomSheet`, for improved user interactions.
	- Integrated a new icon for an empty wallet scenario in the icon library.

- **Changes**
	- Simplified the "Give" and "Find Charity" button functionality for easier user access.
	- Enhanced `ActionTile` usability by making the `isDisabled` state optional.
	- Implemented checks for wallet balance before allowing certain actions, promoting better user guidance. 

- **Bug Fixes**
	- Corrected the interaction flow for action buttons to ensure they are always enabled when appropriate.


Daniela510:
lets not merge before release

lib/features/family/shared/widgets/layout/givt_bottom_sheet.dart
```diff
@@ -0,0 +1,123 @@
...
+  final Widget? headlineContent;
+
+  final GivtElevatedButton? primaryButton;
+  final GivtElevatedSecondaryButton? secondaryButton;
+
+  final VoidCallback? closeAction;
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocBuilder<ImpactGroupsCubit, ImpactGroupsState>(
```
TammiLion, line 32: There's no need for a blocbuilder here as the state is not being used. Where you planning on using the state?

MaikelStuivenberg: nope :) I'll remove that part indeed

lib/features/family/features/profiles/widgets/give_bottomsheet.dart
```diff
@@ -152,4 +171,18 @@ class GiveBottomSheet extends StatelessWidget {
       },
     );
   }
+
+  void showEmptyWalletBottomSheet(BuildContext context) {
+    context.pop();
+
+    showModalBottomSheet<void>(
```
TammiLion, line 178: Except for the context.pop I would make the rest a static method in the EmptyWalletBottomSheet class, so that no matter where you show it, you show it exactly the same and don't have to continuously define the whole "showmodalbottomsheet" blabla. So then you can just do `EmptyWalletBottomSheet.show(context)`

PR #936, opened 2024-08-01
Add option for topup (KIDS-1239)

MaikelStuivenberg:
## Existing screens
image
image

## New screens in this PR
Screenshot 2024-08-01 at 16 48 32
Screenshot 2024-08-01 at 16 49 01
Screenshot 2024-08-01 at 16 48 48
Screenshot 2024-08-01 at 16 48 11

## Summary by CodeRabbit

- **New Features**
	- Introduced a `Topup` functionality, allowing users to manage and execute top-up operations for accounts.
	- Added `TopupWalletBottomSheet` widget, providing a user-friendly interface for wallet top-ups.
	- Implemented `errorCircleWithIcon` widget to enhance error message displays within the app.
	- Enhanced user authentication flow within the `EmptyWalletBottomSheet` for better security in top-up processes.

- **Bug Fixes**
	- Adjusted import paths for `amount_counter` to ensure consistent access across components.

- **Documentation**
	- Updated themes with new error color constants for improved visual feedback during error states.

- **Style**
	- Improved spacing and layout in various UI components for a cleaner user interface.

- **Chores**
	- Cleaned up unnecessary import statements to optimize the codebase.


Daniela510 (approved):

lib/app/app.dart
```diff
@@ -99,6 +100,9 @@ class _AppState extends State<App> {
             create: (BuildContext context) =>
                 ProfilesCubit(getIt(), getIt(), getIt(), getIt(), getIt()),
           ),
+          BlocProvider(
```
TammiLion, line 103: I remember us discussing in the arch meeting that we preferred injecting the cubits via getit rather than relying on them to be passed via context.

lib/features/family/features/topup/cubit/topup_state.dart
```diff
@@ -0,0 +1,30 @@
+part of 'topup_cubit.dart';
+
+enum TopupStatus { initial, loading, done, error }
+
+class TopupState extends Equatable {
+  const TopupState({
+    this.userGuid = '',
+    this.status = TopupStatus.initial,
```
TammiLion, line 8: So personally I wouldn't write the topup state this way.

Right now its implied that every state (initial, loading, done, error) has the same fields: userGuid and error.

https://medium.com/@ssindher11/exploring-sealed-classes-in-flutter-241d3e160132

By using a sealed class instead of an enum you can define per state which fields are present, and whether they are required or not. Fields that are shared you define in the sealed class itself and fields that are specific per state you define in the class that's extending the sealed class.

lib/features/family/features/topup/repository/topup_repository.dart
```diff
@@ -0,0 +1,19 @@
...
+class TopupRepositoryImpl with TopupRepository {
+  TopupRepositoryImpl(
+    this._apiService,
+  );
+
+  final FamilyAPIService _apiService;
+
+  @override
+  Future<bool> topupChild(String childGUID, int amount) async {
+    final response = await _apiService.topUpChild(childGUID, amount);
```
TammiLion, line 16: Currently we have the EditChildRepository and the TopupRepository both doing a topup api call. The first one is doing the v1 call and the 2nd one is doing the v2 call. The first one exposes a stream that updates others that something regarding the child has changed. I don't think this one should expose a different stream because then we lose sight of where to get child data updates.

Do we still need the v1 call? Else i'd say we change the EditChildRepository to use the v2 call and then delete this repo.

PR #938, opened 2024-08-02
Remove all quick flow references (KIDS-1250)

MaikelStuivenberg:


Daniela510 (approved):

lib/features/children/generosity_challenge/pages/generosity_challenge_introduction.dart
```diff
@@ -28,7 +28,6 @@ class GenerosityChallengeIntroduction extends StatefulWidget {
 class _GenerosityChallengeIntroductionState
     extends State<GenerosityChallengeIntroduction> {
   bool _acceptPolicy = false;
-  bool isDebugQuickFlowEnabled = false;
```
Daniela510, line 31: I think we can also remove the GenerosityChallengeCubit from this page and this route since the only reason its here was for the quick flow

PR #950, opened 2024-08-07
Fix: do not display parents on external coin giving

Daniela510:
## Summary by CodeRabbit

- **New Features**
	- Introduced conditional rendering in the profile selection screen based on the flow state, enhancing user experience by showing or hiding components dynamically.
	- Added functionality to reset the flow state during user interactions.

### DIFF
--- a/lib/features/family/features/profiles/screens/profile_selection_screen.dart
+++ b/lib/features/family/features/profiles/screens/profile_selection_screen.dart
@@ -61,6 +61,8 @@ class _ProfileSelectionScreenState extends State<ProfileSelectionScreen> {
       }
     });
     final user = context.read<AuthCubit>().state.user;
+    final flow = context.read<FlowsCubit>().state;
+
     return BlocConsumer<ProfilesCubit, ProfilesState>(
       listener: (context, state) async {
         if (state is ProfilesInvitedToGroup) {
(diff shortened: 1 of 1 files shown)


lib/features/family/features/profiles/screens/profile_selection_screen.dart
```diff
@@ -165,38 +168,39 @@ class _ProfileSelectionScreenState extends State<ProfileSelectionScreen> {
                                 ),
                               ),
                             const SizedBox(height: 8),
-                            GivtElevatedSecondaryButton(
-                              onTap: () async {
-                                if (!context.mounted) return;
+                            if (!flow.isCoin)
```
TammiLion, line 171: I understand that we don't wanna show parents on the external coin flow. Do we also intentionally want to not show the Manage My Family button? Or do we want to give people an "out" like this if they start the app from the external coin flow?

TammiLion (approved):

lib/features/family/features/profiles/screens/profile_selection_screen.dart
Daniela510: yeah you are rught, but i am worried about potential weirdness with the flows, so i will reset the coin flow if they click it

PR #1002, opened 2024-09-02
Remove GC registration screen (KIDS-1329)

MaikelStuivenberg:
Also redirect users directly to registration when needed instead of profile page first

## Summary by CodeRabbit

- **New Features**
	- Enhanced user navigation based on registration status for US users, directing unregistered users to the registration page.
	- Simplified routing structure by removing unnecessary redirection to improve navigation flow.

- **Bug Fixes**
	- Improved navigation error handling by conditionally rendering the back button in the signup page based on the current context.


TammiLion:
did we already remove the first reference to this screen? I only see the path removed from the back button but i was expecting a deletion from the email signup screen as well

lib/features/family/features/profiles/screens/profile_selection_screen.dart
```diff
@@ -92,28 +92,6 @@ class _ProfileSelectionScreenState extends State<ProfileSelectionScreen> {
           } else {
             await context.pushNamed(FamilyPages.childrenOverview.name);
           }
-        } else if (state is ProfilesNeedsRegistration) {
```
MaikelStuivenberg, line 95: This part is not needed because of the fix in approuter and removing the redirection screen

TammiLion: Noice.

TammiLion (approved):

PR #1005, opened 2024-09-03
Fix:  insufficient funds for top up (KIDS-1358)

Daniela510:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new status for member addition indicating "top-up" failures.
	- Enhanced the insufficient funds notification to handle multiple children scenarios with updated messaging and a dismiss button.
  
- **Bug Fixes**
	- Improved error handling during member addition to ensure accurate stream event emissions based on API call outcomes.
  
- **Documentation**
	- Updated comments for clarity regarding event emissions in the member addition process.


lib/features/children/add_member/pages/failed_topup_bottomsheet.dart
```diff
@@ -38,9 +43,11 @@ class InsufficientFundsBottomsheet extends StatelessWidget {
         ],
       ),
       content: BodyMediumText(
-        'We couldn’t take the \$${amount.toStringAsFixed(0)} to top up your child’s wallet.\n\nDon\'t worry we will try again tomorrow.',
+        'We had trouble getting the \$$amount from your account for the $topup.\n\nNo worries, we will try again tomorrow!',
```
TammiLion, line 46: Just to be sure. The amount shown here is the total amount? For example i create 2 children each with a top up of 5. Will $amount then be 10 or will it be 5?

lib/features/children/add_member/repository/add_member_repository.dart
```diff
@@ -25,15 +25,19 @@ class AddMemberRepositoryImpl with AddMemberRepository {
...
-    _memberAddedStreamController.add(null);
-
-    // We add this second event to the stream delayed
-    // because money-related calls require an update from Stripe for the BE
-    // which takes a bit of time
-    Future.delayed(const Duration(seconds: 2), () {
+    try {
+      await apiService.addMember(body);
+    } finally {
+      //does not get emmited if call fails
```
TammiLion, line 31: What do you mean with this comment? Because the finally block will always be executed.

lib/features/children/add_member/widgets/add_member_loading_page.dart
```diff
@@ -29,14 +29,16 @@ class AddMemberLoadingPage extends StatelessWidget {
             SnackBarHelper.showMessage(context, text: state.error);
           } else if (state.status == AddMemberStateStatus.successCached) {
             _navigateToProfileSelection(context);
-          } else if (state.status == AddMemberStateStatus.successNoAllowances) {
```
TammiLion, line 32: We don't need an if case for the successNoAllowance anymore?

TammiLion (approved):
Some small comments to be looked at before merging.

lib/features/children/add_member/repository/add_member_repository.dart
Daniela510: oops, that was the case before, i corrected the comment should say: //always add even if there are errors

lib/features/children/add_member/widgets/add_member_loading_page.dart
Daniela510: i deleted both AddMemberStateStatus.successNoAllowances and AddMemberStateStatus.successNoTopups

lib/features/children/add_member/pages/failed_topup_bottomsheet.dart
Daniela510: Andy decided to reword this without an amount so we dont have to bother with vpc, totals admin fee, it is a nightmare to phrase

PR #1006, opened 2024-09-03
Always logout when going back from registration (KIDS-1360)

MaikelStuivenberg:
## Summary by CodeRabbit

- **New Features**
	- The signup page now includes a `GenerosityBackButton` that always appears, allowing users to log out instead of navigating back.
  
- **User Experience Changes**
	- The back navigation functionality has been replaced with a logout action, altering how users interact with the signup page.

### DIFF
--- a/lib/features/registration/pages/us_signup_page.dart
+++ b/lib/features/registration/pages/us_signup_page.dart
@@ -83,7 +83,7 @@ class _UsSignUpPageState extends State<UsSignUpPage> {
           appBar: GenerosityAppBar(
             leading: GenerosityBackButton(
                     onPressed: () {
-                      logout(context);
+                      logout(context, fromLogoutBtn: true);
                     },
                   ),
             title: 'Enter your details',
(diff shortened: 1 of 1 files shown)


lib/features/registration/pages/us_signup_page.dart
```diff
@@ -80,13 +81,11 @@ class _UsSignUpPageState extends State<UsSignUpPage> {
 
         return FamilyScaffold(
           appBar: GenerosityAppBar(
-            leading: context.canPop()
-                ? GenerosityBackButton(
+            leading: GenerosityBackButton(
                     onPressed: () {
-                      context.pop();
+                      logout(context, fromLogoutBtn: true);
```
TammiLion, line 86: Logging out will cause a redirect in the router?

TammiLion (approved):
Just a question to look at.

lib/features/registration/pages/us_signup_page.dart
MaikelStuivenberg: yes

PR #1077, opened 2024-09-19
Fetch recommendation for profiles (KIDS-1415)

MaikelStuivenberg:


lib/features/family/features/reflect/bloc/grateful_cubit.dart
```diff
@@ -23,77 +24,99 @@ class GratefulCubit extends CommonCubit<GratefulUIModel, GratefulCustom> {
...
+
     if (_profiles.isEmpty) {
       _onEveryoneDonated();
     }
   }
 
   GameProfile _getCurrentProfile() => _profiles[_currentProfileIndex];
 
-  void _emitData() {
+  void _emitData(List<Organisation> currentRecommendations) {
```
TammiLion, line 55: I would make currentRecommendations a field so the emitData does not need any params. Its a method that could be called in more cases than just retrieving recommendations so having to pass it all the time can become quite tedious.

lib/features/family/features/reflect/domain/grateful_recommendations_repository_impl.dart
```diff
@@ -1,28 +1,49 @@
...
   final Map<GameProfile, List<Organisation>> _gratefulRecommendations = {};
 
   @override
   Future<Map<GameProfile, List<Organisation>>>
       getGratefulRecommendationsForMultipleProfiles(
     List<GameProfile> profiles,
   ) async {
-    throw UnimplementedError();
+    final results = await Future.wait(
+      profiles.map((profile) async {
```
TammiLion, line 21: I would either put this in an all-encompasing try catch or the usage of it.
Also if you do it here I would not let it return anything and rename the method like we discussed in the previous PR.

lib/features/family/features/reflect/bloc/grateful_cubit.dart
```diff
@@ -49,50 +54,70 @@ class GratefulCubit extends CommonCubit<GratefulUIModel, GratefulCustom> {
...
         ),
         recommendationsUIModel: null,
       ),
     );
   }
 
-  void onAvatarTapped(int index) {
+  Future<void> onAvatarTapped(int index) async {
     _currentProfileIndex = index;
+
```
TammiLion, line 82: Just add an emitLoading() and should be perfect I think. (is in CommonCubit)

TammiLion: Can also be something for a different PR.

TammiLion (approved):
Approved, added some comments.

lib/features/family/features/reflect/bloc/grateful_cubit.dart
```diff
@@ -49,50 +54,73 @@ class GratefulCubit extends CommonCubit<GratefulUIModel, GratefulCustom> {
...
+          profile: _getCurrentProfile(),
+          organisation: organisation,
+        ),
+      );
     }
   }
 
-  void onDonated(GameProfile profile) {
+  Future<void> onDonated(GameProfile profile) async {
+    emitLoading();
```
TammiLion, line 111: I would only put this in the else. Otherwise you emit the loading. Navigate to next. Can press the back button and be in a perpetual loading state.

lib/features/family/features/reflect/domain/grateful_recommendations_repository_impl.dart
MaikelStuivenberg: Some error handling is added, method renamed and fetching just one as fallback is also improved :)

PR #1078, opened 2024-09-19
UI improvements KIDS-1405

MaikelStuivenberg:


lib/features/family/features/reflect/presentation/pages/gratitude_selection_screen.dart
```diff
@@ -22,10 +24,20 @@ class GratitudeSelectionScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return FunScaffold(
-      appBar: FunTopAppBar.primary99(title: 'Gratitude'),
+      appBar: FunTopAppBar.primary99(
+        title: 'Gratitude',
+        actions: [
+          IconButton(
```
TammiLion, line 30: I would extract this in a widget so it can be re-used across the screens (for example in the GratefulScreen)

TammiLion (approved):
1 small comment

lib/features/family/features/reflect/presentation/pages/gratitude_selection_screen.dart
MaikelStuivenberg: I extracted to button and changed it on all pages were this button is used :)

PR #1080, opened 2024-09-21
Feature: Loop donation flow into reflectio ( KIDS-1417 )

Daniela510:
The pubspec downgrade of firebase was because since @TammiLion PR friday my cocoapods could not resolve between firbase packages and mobile_scanner package. Not sure how it kep working for others, if there is a way to keep the latest package versions lmk 🙏 

## Release Notes

- **New Features**
	- Enhanced navigation to success screens with customizable actions after transactions.
	- Introduced new handling for donation flows, improving user experience during transactions.

- **Bug Fixes**
	- Improved logic for profile recommendations based on donation status.

- **Chores**
	- Cleaned up unused import statements across various screens.
	- Updated dependencies in the project for better performance and compatibility.

### DIFF
--- a/lib/features/family/features/reflect/presentation/pages/grateful_screen.dart
+++ b/lib/features/family/features/reflect/presentation/pages/grateful_screen.dart
@@ -77,8 +77,8 @@ class _GratefulScreenState extends State<GratefulScreen> {
                 child: RecommendationsWidget(
                   uiModel: uiModel.recommendationsUIModel,
                   onRecommendationChosen: (int i) {
-                    context.pop();
                     _cubit.onRecommendationChosen(i);
+                    context.pop();
                   },
                 ),
               ),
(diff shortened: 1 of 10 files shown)


lib/features/family/features/reflect/presentation/pages/grateful_screen.dart
```diff
@@ -62,12 +67,19 @@ class _GratefulScreenState extends State<GratefulScreen> {
             children: [
               GratefulAvatarBar(
                 uiModel: uiModel.avatarBarUIModel,
-                onAvatarTapped: _cubit.onAvatarTapped,
+                onAvatarTapped: (int i) async {
```
TammiLion, line 70: Personally I would revert this change. Then below in handleCustom, the data gives you the profile. And then there I would do the setActiveProfile.

TammiLion (approved):
Nice work! Looks good to me, just one suggestion.

lib/features/family/features/reflect/presentation/pages/grateful_screen.dart
Daniela510: Good idea!

PR #1096, opened 2024-09-24
Feature:  Allow 2 players to play gratitude game (KIDS-1440)

TammiLion:
## Release Notes

- **New Features**
  - The "FunButton" now displays when there are at least 2 profiles instead of 3.
  - Enhanced role management allows multiple roles per profile, improving flexibility in gameplay.
  
- **User Interface Improvements**
  - The "See roles" button is now enabled with at least 2 selected profiles.
  - Updated messaging for role instructions, accommodating multiple roles for clarity.

- **Visual Enhancements**
  - Profiles now display multiple role icons, improving visual representation.

### DIFF
--- a/lib/features/family/features/reflect/domain/reflect_and_share_repository.dart
+++ b/lib/features/family/features/reflect/domain/reflect_and_share_repository.dart
@@ -39,7 +39,7 @@ class ReflectAndShareRepository {
 
   List<GameProfile> getCurrentReporters() {
     return _selectedProfiles
-        .where((profile) => profile.role is Reporter)
+        .where((profile) => profile.roles.whereType<Reporter>().isNotEmpty)
         .toList();
   }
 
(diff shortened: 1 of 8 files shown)


lib/features/family/features/reflect/domain/reflect_and_share_repository.dart
```diff
@@ -119,6 +119,16 @@ class ReflectAndShareRepository {
       role: const Role.sidekick(),
     );
 
+    if (_selectedProfiles.length == 2) {
```
MaikelStuivenberg, line 122: Just personal preference, but it's not really needed to write an 'else' as you just did a return

So just

```
if(..){
return x;
}

return x;
```
could be enough / cleaner

MaikelStuivenberg (approved):

lib/features/family/features/reflect/domain/reflect_and_share_repository.dart
TammiLion: Agreed, will fix

PR #1131, opened 2024-10-04
Bug: Fix rounds navigation and reveal secret word ui fix (KIDS-1498 & KIDS-1467)

Daniela510:
## Summary by CodeRabbit

- **New Features**
	- Introduced conditional navigation in the Pass The Phone feature, allowing users to navigate to different screens based on the new `toRules` parameter.
	- Enhanced the Reveal Secret Word feature by refining the scratching interaction and updating navigation logic.

- **Bug Fixes**
	- Improved text update logic in the Reveal Secret Word screen to ensure it only updates when the scratching state is confirmed.

### DIFF
--- a/lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
+++ b/lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
@@ -28,13 +28,16 @@ class RevealSecretWordScreen extends StatefulWidget {
 class _RevealSecretWordScreenState extends State<RevealSecretWordScreen> {
   final SecretWordCubit _cubit = SecretWordCubit(getIt());
   final scratchKey = GlobalKey<ScratcherState>();
-  bool _isScratched = false;
+  bool _isSecretWordVisible = false;
   bool _isSecondWord = false;
-  String text = 'Scratch to reveal\nyour secret word';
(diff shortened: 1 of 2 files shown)


lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
```diff
@@ -71,7 +71,10 @@ class _RevealSecretWordScreenState extends State<RevealSecretWordScreen> {
                   color: Colors.grey,
                   onChange: (value) => setState(() {
                     _isScratched = value > 25;
-                    text = 'Sneak your secret word\ninto ONE of your answers!';
+                    if (_isScratched) {
```
MaikelStuivenberg, line 74: Why would we change the text in setstates and base the text on _isScratched value? :) 
It's not really clear if the different texts are all over this file

MaikelStuivenberg (changes requested):

lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
Daniela510: i changed the var names to make it clearer

MaikelStuivenberg (approved):

PR #1158, opened 2024-10-10
Disable going back from first time set up family

MaikelStuivenberg:
Maybe not the best solution, but prevents error screen


TammiLion (approved):

TammiLion:
WAIT

PR #1177, opened 2024-10-17
Feature: Change the flow after the first round (KIDS-1560)

TammiLion:


lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
```diff
@@ -107,13 +109,24 @@ class _RevealSecretWordScreenState extends State<RevealSecretWordScreen> {
...
+                      sidekick,
+                      toRules: true,
+                    ).toRoute(context),
+                  );
+                } else {
+                  final reporters = getIt<InterviewCubit>().getReporters();
+                  Navigator.of(context).pushReplacement(
+                    PassThePhone.toReporters(
+                      reporters,
+                      skipRules: true,
```
Daniela510, line 126: confusing that there is both 'toRules' and 'skipRules' for PassThePhone constructors

Daniela510 (approved):

lib/features/family/features/reflect/presentation/pages/reveal_secret_word.dart
TammiLion: True but they are not for the same constructor! One is for the sidekick the other one for the reporters. If it were the same I would not make a double boolean but an enum/ different constructors instead.

PR #1307, opened 2024-12-03
Bug: Color of privacy statement text (KIDS-1695)

Daniela510:
## Release Notes

- **New Features**
	- Enhanced `TermsAndConditionsDialog` to dynamically display terms based on user location.
	- Added customizable title color option to `FunBottomSheet`.

- **Bug Fixes**
	- Corrected repository registration typo in dependency injection setup.
	- Improved error handling in `EmailSignupCubit` for email and country fetching.

- **Refactor**
	- Refactored `TermsAndConditionsDialog` from a `StatelessWidget` to a `StatefulWidget` for better state management.
	- Simplified logging messages in `HomePage` for better clarity.

- **Chores**
	- Removed unused `CountryIsoInfo` registration and related functionality from the project.
	- Updated dependency injection setup to enhance robustness.


lib/features/auth/widgets/terms_and_conditions_dialog.dart
```diff
@@ -1,81 +1,94 @@
...
+}
+
+class _TermsAndConditionsDialogState extends State<TermsAndConditionsDialog> {
+  final _cubit = getIt<EmailSignupCubit>();
+  late String countryIso;
+  @override
+  void initState() {
+    super.initState();
+    _cubit.init();
+    _getCountry();
```
TammiLion, line 31: I would use "then" here because you can't await the async function in the init and you do want to get the country before the widget initially builds considering you are using a late variable. Else you might get a not initialized error if the sharedpreferences take too long.

TammiLion (approved):

lib/features/auth/widgets/terms_and_conditions_dialog.dart
TammiLion: Ignore, i am sleepy, then is not synchronous so it doesn't matter

TammiLion: So I would either already set the country to the same value initially as the cubit (NL) so that even if it builds earlier you don't get an error, or I would reintroduce the futurebuilder

PR #1308, opened 2024-12-03
Bug: Refactor terms dialog to not use TypeOfTerms and pass background darkness (Kids-1695)

Daniela510:


lib/features/auth/widgets/terms_and_conditions_dialog.dart
```diff
@@ -10,11 +10,12 @@ import 'package:go_router/go_router.dart';
...
   const TermsAndConditionsDialog({
-    required this.typeOfTerms,
+    required this.content,
+    this.isDrakBackground = false,
     super.key,
   });
-  //todo refactor this to use the title and the instead of the [TypeOfTerms]
-  final TypeOfTerms typeOfTerms;
+  final String content;
+  final bool isDrakBackground;
```
TammiLion, line 18: typo

lib/features/auth/widgets/terms_and_conditions_dialog.dart
```diff
@@ -44,52 +45,50 @@ class _TermsAndConditionsDialogState extends State<TermsAndConditionsDialog> {
...
-      termsAndConditions = locals.bacsDdGuarantee;
     }
-    if (widget.typeOfTerms == TypeOfTerms.giftAid) {
+    if (widget.content == locals.giftAidInfoBody) {
       title = locals.giftAidInfoTitle;
-      termsAndConditions = locals.giftAidInfoBody;
     }
 
     return Theme(
       data: const FamilyAppTheme().toThemeData(),
```
TammiLion, line 83: I think this is the "original" problem. We should not have the theme object here anymore. Theme switching now occurs upon logging in. Potentially you then don't need the isDarkBackground param anymore. The only place where i'm not sure if it currently works correctly is then the EmailSignupScreen.

Daniela510: no I would still need the isDarkBackground, check giftAid or personal info page, the background is set independent of the theme

Daniela510: But ill remove the theme object anw, you are right its not necessary

TammiLion (approved):

PR #1371, opened 2025-01-02
Bug: Frequent login required when opening the app (KIDS-1785)

TammiLion:
Also added some local fcm push notification testing possibilities. The solution now is that when we get an unauthenticated user update we first check if we can refresh, if that still doesn't work then we actually logout.

## Release Notes

- **New Features**
  - Enhanced notification handling with new test notification scheduling methods
  - Improved authentication initialization error logging

- **Improvements**
  - Refactored home screen cubit to better handle user authentication and session management
  - Added more robust error tracking for authentication and notification processes

- **Bug Fixes**
  - Improved error handling in authentication and notification services

### DIFF
--- a/lib/core/notification/notification_service.dart
+++ b/lib/core/notification/notification_service.dart
@@ -45,6 +45,10 @@ Future<void> _navigateToScreen(NotificationResponse details) async {
     case 'ShowYearlySummary':
       LoggingInfo.instance.info('Navigating to yearly summary screen');
       AppRouter.router.goNamed(Pages.personalSummary.name);
+    case 'FCMTest':
+      LoggingInfo.instance.info('FCM Test');
+      NotificationService.instance
+          .handleFirebaseNotificationNavigation(decodedPayload);
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/core/notification/notification_service.dart
```diff
@@ -139,36 +143,44 @@ class NotificationService implements INotificationService {
     }
   }
 
+  Future<void> scheduleBedtimeTestNotifications() async {
```
Daniela510, line 146: What is this for?

Daniela510 (approved):

lib/features/family/features/home_screen/cubit/family_home_screen_cubit.dart
```diff
@@ -55,6 +53,29 @@ class FamilyHomeScreenCubit
     _emitData();
   }
 
+  Future<void> _handleUserUpdate(UserExt? user) async {
+    if (user == null) {
+      try {
+        final session = await _familyAuthRepository.refreshToken();
+        if (session.isExpired || !session.isLoggedIn) {
+          LoggingInfo.instance.info(
```
Daniela510, line 61: nice 🤞

PR #1435, opened 2025-02-05
Feature: Make the fun xp bar widget (KIDS-1909)

TammiLion:
I need some input from Camila for the animation durations and curves but other than that I'd say it's 99% done.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new customizable progress bar component offering animated visual feedback and multiple styling options.
  - Expanded the app’s theme with an additional neutral color, providing more design versatility.
  - Enabled integration of the progress bar into the family home screen, setting the stage for enhanced visual progress displays.

### DIFF
--- a/lib/features/family/shared/design/components/content/fun_progressbar.dart
+++ b/lib/features/family/shared/design/components/content/fun_progressbar.dart
@@ -2,7 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:givt_app/features/family/shared/widgets/texts/label_small_text.dart';
 import 'package:givt_app/features/family/utils/family_app_theme.dart';
 
-class FunProgressbar extends StatelessWidget {
+class FunProgressbar extends StatefulWidget {
   const FunProgressbar({
     required this.currentProgress,
     required this.total,
@@ -19,6 +19,7 @@ class FunProgressbar extends StatelessWidget {
   factory FunProgressbar.xp({
     required int currentProgress,
     required int total,
+    EdgeInsets? margin,
   }) {
     return FunProgressbar(
       currentProgress: currentProgress,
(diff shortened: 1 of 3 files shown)


lib/features/family/shared/design/components/content/fun_progressbar.dart
```diff
@@ -0,0 +1,148 @@
...
+    required this.total,
+    super.key,
+    this.prefixWidget,
+    this.suffix,
+    this.completedText,
+    this.margin,
+    this.backgroundColor,
+    this.progressColor,
+    this.textColor,
+  });
```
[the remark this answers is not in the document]
TammiLion: Good bot

lib/features/family/shared/design/components/content/fun_progressbar.dart
```diff
@@ -0,0 +1,149 @@
...
+    required this.total,
+    super.key,
+    this.prefixWidget,
+    this.suffix,
+    this.completedText,
+    this.margin,
+    this.backgroundColor,
+    this.progressColor,
+    this.textColor,
+  })  : assert(currentProgress >= 0, 'Current progress must be non-negative'),
```
MaikelStuivenberg, line 17: I didn't know the assert thing :) but it seems like it's only validating that by default on debug versions and not the 'release' version of the app.

Should it be validated there as well? Then we need to add --enable-asserts to the build release command

MaikelStuivenberg (approved):

lib/features/family/shared/design/components/content/fun_progressbar.dart
MaikelStuivenberg: @TammiLion ^^

TammiLion: Good point, considering the values will be coming from the BE I don't think that's particularily useful. I will make sure the widget just hides itself rather than gives an error for the release version.

PR #1656, opened 2025-04-24
Feature: change confetti animation (KIDS-2098)

TammiLion:


lib/shared/dialogs/confetti_dialog.dart
```diff
@@ -1,66 +0,0 @@
...
-        Expanded(
-            child: Align(
-                alignment: Alignment.bottomCenter, child: _confettiAsset())),
-      ],
-    );
-  }
-
-  LottieBuilder _confettiAsset() {
-    return Lottie.asset(
-      'assets/lotties/confetti.json',
```
MaikelStuivenberg, line 61: It should be possible to remove also the file itself, right? :)

MaikelStuivenberg (approved):
see comment

lib/shared/dialogs/confetti_dialog.dart
TammiLion: Yes will do

